PR #33, opened 2016-03-12
Remove peerDependencies

bkniffler:
Many libraries don't define peerDependencies any more (react-router, ..) so we can more freely use react versions of our choice.


mweststrate:
Thanks!

PR #44, opened 2016-04-29
fixes issue with React 15 - React DOM tree root should always have a node reference.

andreigabreanu:
Do not run "forceUpdate()" if the component has *not* been mounted. Fixes Server Side Rendering of a component marked as @observer with React 15.0

### DIFF
--- a/index.js
+++ b/index.js
@@ -46,7 +46,7 @@
                 var isRenderingPending = false;
                 function initialRender() {
                     reaction = new mobx.Reaction(name, function() {
-                        if (!isRenderingPending) {
+                        if (!isRenderingPending && self._isMounted) {
                             isRenderingPending = true;
                             if (typeof self.componentWillReact === "function")
                                 self.componentWillReact();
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks for the PR. I'm not sure however whether this fixes a real problem or that it just hides another one. Would you mind adding a regression test for this change, that fails if this patch isn't applied?

A small code snippet that shows the problem is fine as well.

andreigabreanu:
Agree. I'll provide the regression tests in another PR to expose the issue (if we consider it to be a mobx issue and not a react one).

FYI, the problem occures in other libraries as well - this is the github issue @ react i saw referenced everywhere: https://github.com/facebook/react/issues/6232

Also, from what I understood from the codebase, I see that in componentWillMount mobx-react is adding the callback which calls forceUpdate(). The problem with that - and generally with any code that calls setState() / forceUpdate() in componentWillMount() is that if you render the component server-side (or a unit test) - the component has not mounted but there is a function that can/will be called and will trigger a component update which is incorrect (you can check the explanation in the react issue above).

When I get a few minutes I'll try to add a small code snippet here to illustrate the problem.

mweststrate:
I have to take a closer look, but this starts to ring a bell. I will check later whether this is related, but it reminds me at least of these commits:

https://github.com/mobxjs/mobx-react/commit/7ac924c880747920feaa016a3b4c9c3850449e1a

which I undid again here:

https://github.com/mobxjs/mobx-react/commit/cc2e14e2706e93de2de6da1253a3b20692396d7b

PR #48, opened 2016-05-03
Convert to ES2015 modules

### LINKED ISSUE #43: Rollup error: Module does not export observer
Trying to work mobx-react into one of my projects but I'm having issues getting it to play nicely with Rollup.

     Error: Module c:\Users\cheath\Desktop\front-end\node_modules\mobx-react\index.js does not export observer

Trying to import the lib and decorate a class:

    ....
    import {observer} from "mobx-react";

    @observer
    class Header extends React.Component {
    ....

I can seemingly make it work by making the following mods:

    //mobx-react/index.js
    // Move the export order around.  This project uses Dojo and if the AMD define is before the common we get some issues..
    // UMD
     if (typeof exports === 'object') {
        module.exports = mrFactory(require('mobx'), require('react'), require('react-dom'));
     } else if (typeof define === 'function' && define.amd) {
		define('mobx-react', ['mobx', 'react', 'react-dom'], mrFactory);
     } else {
        this.mobxReact = mrFactory(this['mobx'], this['React'], this['ReactDOM']);
     }

And changing my usage in my component to:

     ....
    import * as MobX from "mobx-react";

    @MobX.default.observer
    class Header extends React.Component {
    ....

Obviously this is a bit of a clunky fix and I am hoping for a better solution.

My Rollup config:

    gulp.task("bundle", function () {
	return rollup.rollup({
		entry: "./src/main.js",
		plugins: [
			nodeResolve({
				main: true,
				skip: [ "esri" ] // Let Dojo handle Esri
			}),
			commonjs(),
			rollupPostcss({
				plugins: [
					require("autoprefixer"),
					require("precss")
				]
			}),
			replace({
				"process.env.NODE_ENV": '"development"'
			}),
			babel({
				exclude: "node_modules/**"
			})
		]
	}).then(function(bundle) {
		bundle.write({
			format: "amd", // AMD is Dojo compatible
			dest: "build/app/main.js",
			sourceMap: true
		});
	})
    });

Nedlinin:
Converted to using ES2015 modules as they are now standard.  

Generated a built version as a CJS file in the dist folder. (Went with CJS rather than UMD as MobX appears to be just CJS.  If you can import MobX you should be able to import this.)

Fixes #43


Nedlinin:
This seems to be a big change but it really isn't that big.  Basically instead of maintaining a UMD module, we switch to using ES2015 modules and harness the power of Babel to build legacy versions if required.

mattruby:
As someone that uses the umd bundles, I'd say it's pretty important.

On Tue, May 3, 2016 at 4:41 PM, Chris Heath <notifications@github.com>
wrote:

> This seems to be a big change but it really isn't that big. Basically
> instead of maintaining a UMD module, we switch to using ES2015 modules and
> harness the power of Babel to build legacy versions if required.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/mobxjs/mobx-react/pull/48#issuecomment-216673546>
>



-- 
-Matt Ruby-
mattruby@gmail.com

Nedlinin:
@mattruby : The goal isn't to get rid of the UMD build but instead to have the source of this module being created using ES2015 modules and have Babel do it's magic to create a CJS module.

I actually overlooked the fact that MobX has a lib/mobx.umd.js file which is (obviously) UMD.

Adding an npm script to build a UMD version and drop it into the dist folder is simple enough if you think this would solve the issue with the PR for you.

Nedlinin:
@mattruby : Now we generate both a UMD and a CJS version.  Would you be willing to try the UMD version and make sure there are no noticeable issues?

mattruby:
I'll try anything once (or twice). I've been working with npm cdn. I should
be able to work with anything.  I just wanted to make sure that everyone
knows that. there are still a few of us rocking es5.

mweststrate:
Thanks for the PR! I need a few days to catch up with all running stuff, but I will definitely take an in depth look at this PR :)

mweststrate:
This PR has some issues currently:

* generated files are checked in, they can be ignored (but not in npmignore!)
* it seems that no separate native / custom builds are build anymore

Although this approach is more modern, I don't think this PR solves any urgent issues at the moment anymore? So I think I'll close it for now until it somebody needs it or fixes they remaining things :)

Thanks for the effort nonetheless!

PR #57, opened 2016-05-21
Fix for React-native v0.26

danieldunderfelt:
Hi!

Since React-Native v0.25 `Component` and `PropTypes` and `React` itself is imported from the normal `react` package, not `react-native`. In v0.26, the `react` stuff was completely removed from `react-native`, leading to errors. This pull request fixes that by importing `react` into mrFactory (great variable name!) instead of `react-native`. The only changed file is the react-native git patch, but I also added the `native.js` file to version control in order to get my own code up and running quickly. As such, this pull request should maybe not be merged directly, but the react-native 0.26 fix itself can be cherry-picked from its commit.


mweststrate:
Merged, thanks! Released as 3.2.0.

PR #60, opened 2016-05-30
Simplified declaration and use of observable React properties, state, and context

phestermcs:
Please see updated API Documentation in README for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,9 +91,97 @@ import {observer} from "mobx-react";
 * `componentWillReact` won't fire before the initial render (use `componentWillMount` instead)
 * `componentWillReact` won't fire when receiving new props or after `setState` calls (use `componentWillUpdate` instead)
 
-### context(propType: React.Validator<any>, defaultValue?: any, nameInContext?: string)
+### Member decorators used to simplify declaring and using observable React properties, state, and context. 
 
-A property decorator that simplifies using observable context. 
+>For clarity, the term 'member' means a javascript property, to distinguish from the term 'property' which means a React Property
+
+These decorators enable accessing React properties, state, and context directly from the component instance (i.e. use `this.myMember` as opposed to `this.props.myMember` or `this.state.myMember` or `this.context.myMember`). Further, the `state` and `context` decorators make their decorated members observable.
+
+* `property` This merely makes a React property accessible directly from the component instance, and throws an exception if an attempt is made to assign.
+* `state` Creates a React property that can be used to easily initialize the state member, and makes the state member observable. After initialization, the state member can be changed, and mobx will kick in.
+* `context` Like the `state` decorator, except uses React's context scoping rules. This same decorator is used by both context providers and consumers, as will be explained below.
+
+>`componentWillReceiveProps` is intercepted to update `state` or `context` members when their related property is set. This should handle most uses cases. However, if more complex logic is required to set state or context based on properties, `componentWillReceiveProps` may still be overridden.
+
+### property(propType: React.Validator<any>, defaultValue?: any)
+
+```javascript
+import {property} from 'mobx-react'
+
+// ---- ES6 syntax
+
+class App extends React.Component {
+    render() {
+        return <MyComponent someProperty="prop init"/>
+    }
+}
+
+@observer 
+class MyComponent extends React.Component {
+    @property(React.PropTypes.string, 'default if not inited')
+    someProperty
+    
+    onClick = () => {
+        // this will throw an exception
+        this.someProperty += ' clicked'
+    }
+    
+    render() {
+        return <div>
+            <div>someProperty = {this.someProperty}</div>
+            <button onClick={this.onClick}>Change</button>
+        </div>
+    }
+}
+
+// ---- ES5 syntax 
+var MyComponent = React.createClass({
+    ...
+})
+// TODO: veryify this works in ES5
+property(React.PropTypes.string, 'default if not inited')(MyComponent.prototype, 'someProperty')
+```
+
+### state(propType: React.Validator<any>, defaultValue?: any)
+
+```javascript
+import {state} from 'mobx-react'
+
+// ---- ES6 syntax
+
+class App extends React.Component {
+    render() {
+        return <MyComponent someState="state init"/>
+    }
+}
+
+@observer 
+class MyComponent extends React.Component {
+    @state(React.PropTypes.string, 'default if not inited')
+    someState
+    
+    onClick = () => {
+        // this will cause mobx to kick in 
+        this.someState += ' clicked'
+    }
+    
+    render() {
+        return <div>
+            <div>someState = {this.someState}</div>
+            <button onClick={this.onClick}>Change</button>
+        </div>
+    }
+}
+
+// ---- ES5 syntax 
+var MyComponent = React.createClass({
+    ...
+})
+// TODO: veryify this works in ES5
+state(React.PropTypes.string, 'default if not inited')(MyComponent.prototype, 'someState')
+```
+
+### context(propType: React.Validator<any>, defaultValue?: any, nameInContext?: string)
 
 ```javascript
 import {context} from 'mobx-react'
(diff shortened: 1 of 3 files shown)


phestermcs:
Seems there zero interest

mweststrate:
Sorry for not responding earlier, I actually missed the notification and didn't see the PR until last week. So apologies for not responding earlier given the effort you have put into it. I superficially looked at it but at first glance the PR seemed to add a lot of new concepts without solving pressing matters (especially since the introduction of `Provider`).

As a tip, before implementing an extensive feature like this, open an issue to discuss the problem, solution and your motiviation first to have an idea whether the PR is likely to get accepted. (You might have noticed I do the same thing regularly, even though its my 'own' project).

phestermcs:
No worries. Was only a day of effort. We're using this functionality regardless, just seemed a good fit for mobx-react (we have it in a separate module atm) and thought others might benefit. Had made some tweaks to our module and having not seen any interest here, decided to close the PR as this fork/branch were created specifically to initiate discussions here (i.e. we were copy/pasting the ts transpiled version into the forked branch just to keep it current with our little module)

If you're still interested in discussing and potentially including in mobx-react, certainly happy to do that.

The initial motivation was to provide a simple means of dealing with observable context, which lead to rounding out with props and state.

The doc changes in the PR provide more detail.

phestermcs:
@mweststrate At the time, prior to `Provider` handling observable context was a pressing matter. Having now looked at `Provider` it seems using the @context decorator is more straight forward, and works better with TypeScript. Curious what specifically you mean by 'lots of new concepts'? We viewed the decorators a simpler means of using existing concepts, while also make things observable.

phestermcs:
Hi @mweststrate, do you have any interest at all in reviewing and discussing the essence of this pull-request? We feel these three simple decorators simplify using mobx with react, and even just using react, and they seem a good fit for mobx-react. If you're already certain this not to be the class, we'd greatly appreciate knowing, so that we may decide on a good home for them.
thx

mweststrate:
Hi @phestermcs I think the most challenging thing in general is to keep things simple and predictable. So hence I am not sure yet whether the proposal for statefull function components like I proposed myself in #69 should be done. More concepts for people new to the library versus more convenience for people experienced with it. 

I quite like your proposal (especially that the context things are strong typed!) but I am worried about two things as well: 
1. It is non idiomatic  React (no `props`, `state` and `context` in between), and I feel I'm already stretching it with defining observables directly on component instances :). However if you don't care too much about "idiomatic" react (and the larger your project / team is, the more consciousnesses pays off against being idiomatic)
2. It are three new decorators for which you have to remember the signature, and which are specific to React components / `mobx-react` (`@observable` instance properties are still generic MobX)

But as said I definitely see how your extension is convenient, especially in combination with TypeScript. Would you mind publishing it as a separate package? I gladly promote it and then we can just see how this is picked up by the community. Would that be a feasible thing to do for you? If there is general consensus about this being nicer & more convenient, we can always provide it out of the box with `mobx-react` (that is actually how `Provider` for example came into the project).

phestermcs:
@mweststrate, Thanks kindly for responding. I think there's some misunderstanding here as to how these decorates actually work, and which common patterns they greatly simplify while implementing idiomatically. Please read this with an open mind, and an attention to detail, and reconsider incorporating into mobx-react. If you still feel they don't fit, or are somehow not improving 'idiomatic' use of React, then we will put them somewhere else.

>  It is non idiomatic React (no props, state and context in between)

Actually, these three decorates do in fact wrap `props`, `state`, and `context`, so using them both simplifies and improves adherence to using 'idiomatic' React when it comes to `props`, `state`, and `context`, but does so in a way that fully embraces the power and simplicity of using mobx.

> It are three new decorators for which you have to remember the signature, and which are specific to React components / mobx-react (@observable instance properties are still generic MobX)

First, if having to remember three signatures is a substantially inhibiting burden for the programmer, they really should be in another line of work, and in the simplest use of the decorators, there is no signature to remember; the couple of additional and optional arguments are to further simply working with `props`, `state`, and `context` in the idiomatic way, while embracing mobx. Second, your are exactly correct these decorators are specific to using React /mobx-react, which is why we strongly believe they belong with mobx-react.

The intent of these three decorators is specifically to make things simpler and more prescriptive when using React and mobx together. Certainly, someone new to React will already have much to learn, and upon learning, will look for ways to simplify.

In my humble opinion, the moment you embrace using mobx with React, you're taking a very explicit step away from FB's definition of 'idiomatic' React, in that mobx's FRP handles change detection in an extremely simple (in use, via mobx-react), understandable, and efficient way compared to the 'idiomatic' Flux / Store / Immutable / `shouldComponentUpdate()` way. 

What lead us to mobx (which is freakin' awesome btw), was the fair amount of complexity in optimizing `shouldComponentUpdate()` especially considering descendant `shouldComponentUpdates()` are not called if their parent `shouldComponentUpdate()` returns false, and optimization is required for 'real' applications that are much more than the TodoMvc sort; i.e. having deep and complex domain models (dozens to hundreds of deeply related entity types), and corresponding UI trees.

Consider these common patterns, and then how they're simplified using these decorators along with mobx & mobx-react (note all code is ES6):

### Define a property that's used to initialize some state

_**The less simple, less predictable, less performant way:**_

```javascript
class MyComponent extends React.Component {
	static propTypes = {
		isEnabled: React.PropTypes.bool
	}
	static defaultProps = {
		isEnabled: true
	}
	state = {
		isEnabled: this.props.isEnabled
	}
	shouldComponentUpdate(nextProps, nextState) {
		// if not using immutable, deep compare hurts
		return nextState.isEnabled !== this.state.isEnabled			
	}
}
```
_**The simple and predictable way, identical to above**:_ 
(i.e. `@state` is configuring `propTypes`, `defaultProps`, and `this.state` 'idiomatically', just like the above, but `isEnabled` is `@observable`)
```javascript
@observer
class MyComponent extends React.Component {
	@state(React.PropsTypes.bool, true) isEnabled
}
```
Given that `propTypes` and `defaultProps` are still being set by the decorator, React tooling still works, and because this  is done much more simply, it increases the likelihood that `propTypes` and `defaultProps` are actually being configured correctly, or 'idiomatically'.

But even better, the decorator also configures `isEnabled` as an observable property within `this.state`, so things like react-hot-loader w/webpack can still detach and reattach the state correctly, which is hard to do using `@observable` directly. 

Lastly, an `isEnabled` getter/setter off the instance provides convenient (much less code verbosity) access to `this.state.isEnabled`

You can still do this:
```javascript
@observer
class MyComponent extends React.Component {
	@state isEnabled
}
```
Which would just make `this.state.isEnabled` observable, while providing the `isEnabled` getter/setter access to it

And this:
```javascript
@observer
class MyComponent extends React.Component {
	@state(true) isEnabled
}
```
Whic sets the `isEnabled` to the default, but doesn't create an actual React property (i.e. `propTypes` and `defaultProps` aren't set.

### Define a property that's used to initialize observable context

_**The way, way, way less simple, less predictable, less performant way:**_
```javascript
class ParentContext extends React.Component {
	static propTypes = {
		isEnabled: React.PropTypes.bool
	}
	static defaultProps = {
		isEnabled: true
	}
	static childContextTypes {
		isEnabled: React.PropTypes.bool
	}
	state = {
		isEnabled: this.props.isEnabled
	}
	getChildContext {
		return {
			isEnabled: this.state.isEnabled
		}
	}
	shouldComponentUpdate(nextProps, nextState) {
		// if not using immutable, deep compare hurts
		return nextState.isEnabled !== this.state.isEnabled			
	}

	onClick = () => {
		this.setState({isEnabled: !this.state.isEnabled])
	}

	render() { 
		return <div>
			<button onClick={this.onClick}>Change Context</button>
			<ChildUsingContext/> 
		</div>
	}
}

class ChildUsingContext extends React.Component {
	static contextTypes {
		isEnabled: React.PropTypes.bool
	}

	shouldComponentUpdate(nextProps, nextState, nextContext) {
		// deep compare hurts, about as much as programming with 
		// immutable everywhere in attempts to avoid it
		return nextContext.isEnabled !== this.context.isEnabled
	}

	render() {
		// ugly way to "default" context.isEnabled
		return <div>{this.context.isEnabled || true}</div> 
	}
}
```

Note, the above still suffers from [react issue #2517](https://github.com/facebook/react/issues/2517), and prior to `Provider` or the `@context` decorator, there was no way to actually have `@observable` context.

_**The super simple, performant way:**_
(but still idiomatically identical to above in that `this.context` is used, along with `propTypes` and `defaultProps` being set correctly, doesn't suffer issue #2517, and enables actually having something off of `this.context` be `@observable`)
```javascript
class ParentContext extends React.Component {
	@context(React.PropTypes.bool, false) isEnabled
	render() { 
		return <div>
			<button onClick={this.isEnabled = !this.isEnabled}>Change Context</button>
			<ChildUsingContext/> 
		</div>
	}
}

@observer
class ChildUsingContext extends React.Component {
	@context(true) isEnabled
	render() { return <div>{this.isEnabled}</div> }
}
```

### Define a property
(to round out consistency in using decorators to define React render() data dependencies)
_**The mildy awkward way:**_
```javascript
class MyComponent extends React.Component {
	static propTypes = {
		isEnabled: React.PropTypes.bool
	}
	static defaultProps = {
		isEnabled: true
	}
}
```

_**The less awkward way:**_
(but still identical to above in that `propTypes` and `defaultProps` set correctly, and having convienence getter on instance (i.e. `this.isEnabled` same as `this.props.isEnabled), and setter that throws (to ensure props not being accidently set)
```javascript
class MyComponent extends React.Componet {
	@property(React.PropTypes.bool, true) isEnabled
}
```

mweststrate:
Hi @phestermcs,

I have no problem in seeing the value, elegance and convenience of your decorators. Especially since I am a huge TypeScript fan myself and this definitely helps in that area. I think your examples are really nice and clean and offer a clear value proposition. 

Nonetheless I would prefer to see this as a separate package. Mainly for a few reasons. As said, although this is nicer, it is also less idiomatic and I think that an opt-in strategy is better suited by a separate package. Secondly I really want to keep mobx-react simple, as people use it as basis for binding for other component libraries as well, such as inferno, deku etc etc. The philosophy of the mobx-react package is to have MobX working with React easily, but not to improve React in general. Finally from maintenance perspective I would rather see this as separate package, especially because it relies on decorators, which often brings a lot of additional work with them. What about ES5? Or field initializers? I think these questions should be answered as well, but with a separate package you can simply choose to only support TypeScript, or ESNext, etc.

I have however no problem to make your package a part of the `mobxjs` github organization and adding an section to the official docs. Personally I would call your package something like `mobx-react-decorators` or `awesome-mobx-react` :). I hope you can live with that, because I do like the features!

Sorry for the late response again, it went of my radar as the PR was already closed.

Issue #53, opened 2016-05-10
Built in support for connecting stores to components


arahansen:
A common pattern I see arising is the need to connect a component to one or many mobx stores. 

Currently, you need to write your own higher order component that will fetch these stores. While this isn't a monumental task in itself (as mobx makes this type of thing relatively straight forward), it would be great to have this feature built into mobx-react to promote a standardized method to accomplish this.

Looking to react-redux for inspiration, I see a lot of ideas there that could gracefully transfer over to mobx-react.

One approach is to wrap the application's Root component in a higher order store provider component that will inject application stores into the application's context. Then retrieve stores with an `@connect` (or similar) decorator that defines which stores should be connected.

Another approach is to "register" stores with react-mobx at their definition. Then, use a similar `@connect` decorator as described above to fetch one or several registered stores.

I would like to start a discussion as to why this does/doesn't make sense for mobx-react. And if it does make sense, get other thoughts on the table.

kuuup:
Take a look at react-tunnel. It is inspired by redux and a great match for mobx.

mweststrate:
See pull request #65 , build `3.4.0-beta.1` can be used to test and give feedback on the new Provider mechanism. Docs: https://github.com/mobxjs/mobx-react/blob/provider/README.md#provider-experimental

mweststrate:
Released as 3.4.0!

PR #65, opened 2016-06-21
Introducing Provider, implements #53

mweststrate:
MobX can now pass stores around via context! Use `Provider` to provide them, and `@observer([store names])` to make them available in your component's props

Beta build is available: `3.4.0-beta.1`
Docs: https://github.com/mobxjs/mobx-react/blob/provider/README.md#provider-experimental

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,9 @@ import {observer} from "mobx-react";
 * `componentWillReact` won't fire before the initial render (use `componentWillMount` instead)
 * `componentWillReact` won't fire when receiving new props or after `setState` calls (use `componentWillUpdate` instead)
 
-### `Provider`
+### `Provider` (Experimental)
+
+_This feature is marked as experimental as the exact api might change in a next minor, pending any community feedback_.
 
 `Provider` is a component that can stores (or other stuff) on React's context.
 This is useful if you have things that you don't want to pass through multiple layers of components explicitly.
@@ -139,7 +141,7 @@ class MessageList extends React.Component {
 
 Some note about passing stores around:
 * If a component ask a store and receives a store via a property with the same name, the property takes precedence. Use this to your advantage when testing!
-* Values provided through `Provider` should be final, to avoid issues like 
+* Values provided through `Provider` should be final, to avoid issues like mentioned in [React #2517](https://github.com/facebook/react/issues/2517) and [React #3973](https://github.com/facebook/react/pull/3973), where optimizations might stop the propagation of new context. Instead, make sure that if you put things in `context` that might change over time, that they are `@observable` or provide some other means to listen to changes, like callbacks.
 
 ## FAQ
 
--- a/README.md
+++ b/README.md
@@ -91,6 +91,56 @@ import {observer} from "mobx-react";
 * `componentWillReact` won't fire before the initial render (use `componentWillMount` instead)
 * `componentWillReact` won't fire when receiving new props or after `setState` calls (use `componentWillUpdate` instead)
 
+### `Provider`
+
+`Provider` is a component that can stores (or other stuff) on React's context.
+This is useful if you have things that you don't want to pass through multiple layers of components explicitly.
+By passing a string array as first argument to `observer`, observer will pick up the named stores from the context and make them available as props of de decorated component:
+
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -175,40 +228,83 @@
...
+                            return;
+                        if (!(storeName in baseStores))
+                            throw new Error("MobX observer: Store '" + storeName + "' is not available! Make sure it is provided by some Provider");
+                        newProps[storeName] = baseStores[storeName];
+                    }, this);
+                    return React.createElement(component, newProps);
+                }
             });
-        };
+            Injector.contextTypes = { mobxStores: PropTypes.object.isRequired };
```
mweststrate: Should not be required but optional, because it is allowed to provide stores using props instead of context (for example in tests this is useful)

royriojas:
Very nice. This is just to get a value from the context right? But that value can be a mobx instance or not right? Just curious.

mweststrate:
@royriojas yes can be anything. But in combination with MobX powered stores it is really powerful, as `observer` components will then automatically reflect anything that happens in those context stores nicely.

README.md
```diff
@@ -91,6 +91,58 @@ import {observer} from "mobx-react";
 * `componentWillReact` won't fire before the initial render (use `componentWillMount` instead)
 * `componentWillReact` won't fire when receiving new props or after `setState` calls (use `componentWillUpdate` instead)
 
+### `Provider` (Experimental)
+
+_This feature is marked as experimental as the exact api might change in a next minor, pending any community feedback_.
+
+`Provider` is a component that can stores (or other stuff) on React's context.
```
winkler1: "can stores"?  Seems to be missing a verb. Perhaps "can pass stores"?

README.md
```diff
@@ -91,6 +91,58 @@ import {observer} from "mobx-react";
...
+This is useful if you have things that you don't want to pass through multiple layers of components explicitly.
+By passing a string array as first argument to `observer`, observer will pick up the named stores from the context and make them available as props of de decorated component:
```
winkler1: "de" -> "the"

ericclemmons:
First, thank you for this PR. It's a very valuable piece that's a mild annoyance wiring up, compared to how little boilerplate MobX requires.

- How do collisions work? As in, my component uses both `foreground.color` and `background.color`?

Because of this, would an object reference be better?

```js
@observer({
 fColor: "foreground.color",
 bColor: "foreground.color",
})
```

Forgive me if I missed something. I'm in Disneyland Paris checking up on tech :)

antitoxic:
I understand some people got used to having `<Provider>` components, but I wonder how is this better than simply using React's ContextTypes? Having defined contextTypes, a function similar to redux' `@connect` is trival.

mweststrate:
@ericclemmons I think it is a bit simpler than that, provider just puts literally the same things in  using components props as you put into the provider. If providers are nested, the inner one wins.

So you could do something like:

```javascript
const colors = observable({
   foreground: '#000',
   background: '#fff'
})

const App = () => {
  <Provider colors={colors}>
     <app stuff... />
  </Provider>
}

const Button = observer(["colors"], ({ colors, label, onClick }) => 
  <button style={{
      color: colors.foreground,
      backgroundColor: colors.background
    }}
    onClick={onClick}
  >{label}<button>
)

// later..
colors.foreground = 'blue';
// all buttons updated

```

Does that answer the question?

mweststrate:
@antitoxic It just avoids some boilerplate around defining `contextTypes` on producer and consumer, created a `getChildContext` and has clearly defined semantics and checks around the presence of stores. This just makes using context simpler (e.g. hopefully you don't need the manual each time), not different. It's a convention over configuration approach

arahansen:
@mweststrate what are your thoughts on tweaking the API to take an arbitrary number of parameters instead of an array?

example: `@observer('colors', 'posts', 'users')`

It would be a simpler interface instead of the extra syntax introduced using an array.

mweststrate:
@arahansen yes, thought about that, but it might introduce hard to disambiguate scenarios in the future. I can imagine for example introducing a first `displayName` parameter at some point.

PR #67, opened 2016-06-22
Fixed #66 - Doesn't work in IE 10

### LINKED ISSUE #66: Doesn't work in IE 10
The reportRendering method use WeakMap which is not available in IE 10 and the method fails. Preparing simple PR for that.

pvasek:
I just added checks before using the componentByNodeRegistry variable. Not sure how devtools will handle the case that it's undefined.

### DIFF
--- a/index.js
+++ b/index.js
@@ -18,7 +18,7 @@
 
         function reportRendering(component) {
             var node = findDOMNode(component);
-            if (node)
+            if (node && componentByNodeRegistery)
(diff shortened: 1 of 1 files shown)


pvasek:
It seems that the devtools handles it correctly with message:
_[mobx-react] tracking components is not supported in this browser._

mweststrate:
Thanks!

Issue #71, opened 2016-06-30
Provider doesn't seem to work in an isomorphic setting 


prashaantt:
I use a static method (the typical `fetchData()`) in my components to load all the required data before rendering on the server. Using the `Provider` API no longer gives me access to that static method.

prashaantt:
To be more specific, the static method isn't available any longer if I use the `observer(["store"])` construct. Using `Provider` by itself isn't the real reason.

mattruby:
I think I can see what you're up against here.  The observer with store is calling createStoreInjector which calls React.createClass with the correct context settings.  You component is wrapped up in the render call of the wrapper.  Hmmmm...  Give me a few to look into what could be done to attach your statics to the wrapping component.  Can you think of anything other than statics that would need to be applied to the wrapper?

mattruby:
@prashaantt Give my PR a look.  See if that fixes the issue with statics.

prashaantt:
Thanks, this works perfectly!

I can't think of anything else for the moment that Provider should surface, but will surely report back if I do.

mweststrate:
@prashaantt didn't merge the proposed PR. I think it is a bit fuzzy and unidiomatic. It is better to either separate the inner and higher order component. Also the original component is now available on the higher order component as `.wrappedComponent`. If anything is unclear, feel free to re-open this issue.

prashaantt:
@mweststrate The good thing is, I'm not sure if you intended it but I now have access to my `fetchData` inside the `wrappedComponent` all right!

mattruby:
@prashaantt pretty sure that was @mweststrate 's intent.  His wrappedComponent is the better way to go.

Issue #70, opened 2016-06-23
Prop validation throws a warning when used on a decorated component expecting store


dmi-dmi-dmi:
When I run the code below I get `Warning: Failed propType: Required prop 'store' was not specified in 'MobXStoreInjector'. Check the render method of 'Root'` warning in the browser console.

```
import React, { Component, PropTypes } from 'react';
import ReactDOM from 'react-dom';
import { Provider, observer } from 'mobx-react';


const store = {};

@observer(['store']) class MyComponent extends Component {

  render() {
    return <div>abc</div>;
  }

}

MyComponent.propTypes = {
  store: PropTypes.object.isRequired,
};


function Root({ store }) {
  return (
    <Provider store={store}>
      <MyComponent />
    </Provider>
  );
}

ReactDOM.render(
    <Root store={store} />,
  document.getElementById('reactApp'));
```

mweststrate:
Quick question, do you use Typescript or Babel for transpiling?

dmi-dmi-dmi:
Babel 6

jtraub:
I've stumbled across this problem today.

As far as I understand it is not a bug at all.

If you read [decorator proposal](https://github.com/wycats/javascript-decorators) it will become clear that 

```
@observer(['store']) class MyComponent {
...
}
```
is equivalent to 

```
let MyComponent = (function () {
  class MyComponent {
  ...
  }
  
  return observer(['store])(MyComponent) || MyComponent
})();
```

At this line 
```
MyComponent.propTypes = {
``` 
we are attaching propTypes object to `MobXStoreInjector` instead of `MyComponent`. Hence, we get warning from React regarding required store prop: it doesn't receive such prop.

To confirm this let's inspect what Babel 6 without es2015 preset produces from the code in the first message:

```
import React, { Component, PropTypes } from 'react';
import ReactDOM from 'react-dom';
import { Provider, observer } from 'mobx-react';

const store = {};

let MyComponent = (_dec = observer(['store']), _dec(_class = class MyComponent extends Component {

  render() {
    return React.createElement(
      'div',
      null,
      'abc'
    );
  }

}) || _class);


MyComponent.propTypes = {
  store: PropTypes.object.isRequired
};

function Root({ store }) {
  return React.createElement(
    Provider,
    { store: store },
    React.createElement(MyComponent, null)
  );
}

ReactDOM.render(React.createElement(Root, { store: store }), document.getElementById('reactApp'));
```

[...]

PR #72, opened 2016-06-30
#71 Added statics support to Injector components.

mattruby:
Provider doesn't seem to work in an isomorphic setting #71
I'm not sure if I'm pulling the potential static methods in the best way.

### DIFF
--- a/index.js
+++ b/index.js
@@ -278,8 +278,17 @@
          * Store Injection
          */
         function createStoreInjector(stores, component) {
+            
+            var statics = Object.getOwnPropertyNames(component).filter(function (propName) {
+				return typeof component[propName] === 'function';
+			}).reduce(function (staticMethods, staticName) {
+				staticMethods[staticName] = component[staticName];
+				return staticMethods;
+			}, {});
+			
             var Injector = React.createClass({
                 displayName: "MobXStoreInjector",
+                statics: statics,
                 render: function() {
                     var newProps = {};
                     for (var key in this.props)
--- a/test/context.js
+++ b/test/context.js
@@ -174,7 +174,7 @@ test('warning is printed when changing stores', t => {
     t.end();
 })
 
-test('statics are copied over to the Injector wrapped component', t = > {
+test('statics are copied over to the Injector wrapped component', t => {
 
 	var C = observer(["foo"], React.createClass({
 		statics: {
--- a/test/context.js
+++ b/test/context.js
@@ -173,3 +173,33 @@ test('warning is printed when changing stores', t => {
     console.warn = baseWarn;
     t.end();
 })
+
+test('statics are copied over to the Injector wrapped component', t = > {
+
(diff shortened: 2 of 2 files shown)


prashaantt:
@mattruby @mweststrate Any idea when this can be merged? I jumped early to test and implement the new Provider (which works great) and now my current project wholly depends on this being available. :)

mweststrate:
@prashaantt @mattruby I'm not entirely sure yet whether this will be merged, the proposed solution in #77, separating `observer` into `observer` and a separate HoC seems to be an more idiomatic React approach to these kind of problems. Would that work as well for you? 

Btw, I think publishing static methods on a component is also a bit uncommon thing to do, why not just export the static method as normal method? But alas, would the following patterns solve your problem as well?

```javascript
@observer MyClassImpl extends React.Component {

}

export const MyClass = @inject("store")(MyClassImpl)
MyClass.fetchData = function() { .. }
```

prashaantt:
> Would that work as well for you?

Sure, that should work just as well. Thanks, looking forward to #77!

> would the following patterns solve your problem as well?

Your example as given doesn't compile in TypeScript. I have [listed here](http://www.typescriptlang.org/play/index.html#src=class%20MyComponent%20%7B%0A%09static%20staticFunc()%20%7B%7D%0A%09instanceFunc()%20%7B%7D%0A%7D%0A%0AMyComponent.staticFunc()%3B%20%2F%2F%20yay%0A%0AMyComponent.prototype.instanceFunc()%3B%20%2F%2F%20ok%0A%0AMyComponent.otherFunc%20%3D%20()%20%3D%3E%20%7B%7D%3B%20%2F%2F%20nay) a few ways to attempt this. Defining the `fetchData` on the component's prototype is an option, but the problem with that (even with the current implementation of `@observer(["store"])` is that my actual component gets buried under several layers of decorated functions and it's hard -- if not impossible -- to get to the prototype of the component I'm actually looking for in order to call the method. I didn't see it this way before, but this perhaps validates why static methods are actually the recommended way of achieving isomorphic rendering in most online articles.

index.js
```diff
@@ -278,8 +278,17 @@
          * Store Injection
          */
         function createStoreInjector(stores, component) {
+            
+            var statics = Object.getOwnPropertyNames(component).filter(function (propName) {
+				return typeof component[propName] === 'function';
```
prashaantt: @mattruby Just wanted to give a heads up, this fails in Safari because `Object.getOwnPropertyNames` apparently doesn't respect strict mode and returns `arguments` and `caller` also as properties. It would perhaps have been better to use `Object.keys` here.

mattruby: Good to know about the getOwnProperties issues.

mweststrate:
Next release will introduce the `wrappedComponent` property on the higher order component to access the original component. By separating `observer` into `observer` and `inject` it is also simpler to access the intermediate object. See also the comments #70 . I hope that helps!

PR #93, opened 2016-07-30
Code split & Better propTypes

andykog:
Added `arrayOrObservableArrayOf` and `observableArrayOf`


index.js
```diff
@@ -342,31 +342,133 @@
...
+
+            // Fallback for non-spec compliant Symbols which are polyfilled.
+            if (typeof Symbol === 'function' && propValue instanceof Symbol) {
+                return true;
+            }
+
+            return false;
+        }
+
+        // Copied from React.PropTypes
```
mweststrate: just checking, are those exposed by react? because react is available here

andykog: Of course no :-)

andykog: I mean, of course they are not exposed. Well, we could leave without isSymbol, getPreciseType if the weight is important. It's sad that the lib can't be split into files for now.

andykog: Are you ok about building it with browserify? That would enable us to use https://github.com/mridgway/hoist-non-react-statics for #92

mweststrate: No problem, but mind the builds for RN / custom react renderers

Op za 30 jul. 2016 om 20:51 schreef Andy Kogut <notifications@github.com>:

andykog: Yeap, I think, that can be done less hacky :-)

andykog:
Splitted code into different files

andykog:
Splitted code into different files
@mweststrate, need your review :-)

package.json
```diff
@@ -36,7 +40,9 @@
     "tape-run": "2.1.0",
     "typescript": "^1.8.10"
   },
-  "dependencies": {},
+  "dependencies": {
+    "webpack": "^1.13.1"
```
mweststrate: dev dependency?

mweststrate:
@andykog looking good to me! Feel free to merge and release whenever you feel confident. Might it be useful to add a minified build for the standard build? The lib has grown quite a bit since introducing Provider and propTypes.

andykog:
> Might it be useful to add a minified build for the standard build?

Yes, I've just temporary turned minification off so it easier to check output

package.json
andykog: Yes, I'll fix that :-)

PR #103, opened 2016-08-11
Pass 'secret' PropType argument to disable warning

jesstelford:
As per the react docs for libraries:
https://facebook.github.io/react/warnings/dont-call-proptypes.html#fixing-the-false-positive-in-third-party-proptypes

Note that I was unable to compile / test this. I had all sorts of issues with `npm link`ing it, and it then failing on `thing instanceof ObservableArray`. In the end, I hacked this into the already-compiled source of the `npm install`d version in my repo, and verified it still worked.

Please test this thoroughly with your normal build process and react@latest.


mweststrate:
Nasty :-P. Thanks for this PR!

PR #117, opened 2016-09-05
Export propTypes as PropTypes

andykog:
I think, that's more intuitive for react users?


mweststrate:
Ha, yeah definitely :)

PR #119, opened 2016-09-06
Fix PropTypes import in Readme.md

reisub-de:
The package is called mobx-react, or is this due to change?
The PropTypes should then be imported from mobx-react and not react-mobx.


andykog:
@reisub-de, thanks :-)

Issue #127, opened 2016-09-19
Webpack's UglifyJsPlugin causing problems.


birkir:
I am having problems with the package when compiling for production. I get the following error:
```bash
s.default.createClass is not a function
```
It is originating from https://github.com/mobxjs/mobx-react/blob/master/src/inject.js#L8 where `React` is uglified to `s.default`.

I can resolve it by either:
1. Build the `mobx-react` package without the **UglifyJsPlugin** in `webpack.config.js` and use it directly in my project.
2. Or use the npm version and build my project without the **UglifyJsPlugin** in `webpack.config.js`.

Questions:
1. Is uglifying the build really necessary? 
2. I cannot reproduce the problem with a dead simple project, as something happens when the code size or dependency tree gets to X size.

birkir:
Ok I've been looking better into this issue. To fix the build, I've added the following to my UglifyJs config:

```js
client.plugins.push(new webpack.optimize.UglifyJsPlugin({
  compress: {
    warnings: true,
  },
  mangle: {
    except: ['r'], // I'm guess r is require.
  },
}));
```

But this is only temp solution because the next version that get's published could generate other variable name for `require`.

Looking into other packages (like `mobx` for example), they don't ship their build uglified. Some packages do it as well but strictly name them `package.min.js`. Also you should check out how many issues the uglifyjs package has.

Thank you.

birkir:
I want to stress this issue again.

Check out the components that are part of the `mobx-react` or `mobx-react-devtools` packages:

![image](https://cloud.githubusercontent.com/assets/180773/18845540/93489d36-8411-11e6-8ef8-9a0badef98e1.png)

This is in development mode. The build phase really needs modification asap.

mweststrate:
@birkir can you share your babel setup (I now it has some plugins for alternative handling of default imports etc)? which babel version?

Anybody having the same issue?

birkir:
Sure, I'm working on the ueno. starter kit: https://github.com/ueno-llc/starter-kit

mweststrate:
@birkir especially your first problem is very weird, this setup is here for half a year and used by many people. Even though uglified require calls to React should be resolved by your webpack setup. Your build setup is quite complicated, would you be able to boil it down to a minimal build setup?

I guess your mangle setup should exclude `require`, not `r`? Does the problem still occur without uglify on your site?

@andykog any ideas?

andykog:
[...]

PR #128, opened 2016-09-26
Build non-minified/non-uglified for npm

birkir:
See #127


mweststrate:
On hold till resolving #127. (But yeah, probably uglify doesn't add much value indeed, maybe there should be even a way to include it unbundled for better build optimizations)

birkir:
Understood.

I am not familiar with packages for npm written in typescript, but I do know that react-redux uses just babel to transpile es6 into es5 that nodejs can use.

They also ship two UMD modules built with webpack. (dev build and uglified production build).

PR #133, opened 2016-10-14
Add React as a peer dependency

### LINKED ISSUE #132: React should be a peer dependency

stevenjob:
Closes #132


mweststrate:
This might be a potential issue if people use this package directly with react derivates (inferno / preact). But we should find that out quickly enough. Note that react 0.13 is also supported, could you add that to your PR, and add a line to the `changelog.md`? Thanks!

PR #134, opened 2016-10-14
Dispose reaction on error inside rendering

andykog:
@mweststrate, sometimes in complex apps, when error happens inside rendering, we can't see original error, because of additional forceUpdate.
I've spent an hour trying to isolate this user-case: https://jsfiddle.net/Sl1v3r/1adco8hx/
May be you have a better idea on how to fix this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: node_js
-script: npm test
+script: npm run travis
 after_success:
   - cat ./coverage/lcov.info|./node_modules/coveralls/bin/coveralls.js
 node_js:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # MobX-React Changelog
 
+### 3.5.9
+
+* Introduced `useStaticRendering(boolean)` to better support server-side rendering scenerios. See [#140](https://github.com/mobxjs/mobx-react/issues/140)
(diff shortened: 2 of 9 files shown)


mweststrate:
Updated the fiddle with a build based on this PR: https://jsfiddle.net/1adco8hx/4/

Can you elaborate why this PR makes a difference? Is that because how React handles errors internally? If that is the case I think this approach is fine. 

Maybe a safety check if `dispose` isn't undefined, although I think theoretically that cannot be the case.

andykog:
@mweststrate, I'm not completely sure, but I think, the problem is that when exception occurs inside render, react have no problem with handling it, but if there is synchronous `forceUpdate` after that, react throws it's own exception in some cases (Cannot read property 'getHostNode' of null). I can try to investigate, maybe that can should be fixed inside React.

mweststrate:
@andykog would you be able to add tests that make the difference clear? I think this is a good addition, but might be nice to see the exact difference

andykog:
@mweststrate, I've found that there is definitely nothing wrong with mobx-react, sometimes react just throws own exception when you call forceUpdate on component that met error during rendering. Here is an example without mobx: https://jsfiddle.net/Sl1v3r/70rd04yp/

So I've simply checked that components stop updating after error: 	48319cc

(Maybe I'l make an issue/pr in react, if I'll manage to understand the root cause.)

mweststrate:
Merged into `v4` branch

Issue #124, opened 2016-09-13
Observable props of @observer-ed component


Strate:
As for now, `this.props` object itself is not an observable, which is a problem if we trying to use it inside `computed` function or bount child component. Some examples:

```
@obeserver
class Component extends React.Component {
  @computed
  get calculatedValue() {
    return this.props.value1 + this.props.value2
  }
}
```

`claculatedValue` does not tracks `this.props` properties, because it is not reactive. And the developer ran into unobviously issue, "why mu computed function does not recompute?"

Second one is bit more complex, but has almost same behaviour:

```
@obeserver
class Component extends React.Component {
  childComponent = observer(props =>
    <div onClick={this.props.onChildClick} />
  )

  render() {
    return <this.childComponent />
  }
}
```

If `Component` receives new prop `onChildClick`, this function is not assigned to child div, because child component does not rerender.

As for now, I'm working on custom decorator, which made `this.props` of component observable. I'll show you my results, and we'll have a little discuss about it :)

mweststrate:
Yeah I definitely see why you want this :)

I can see three options
1) another decorator which makes `this.props` observable (downside: another decorator)
2) by default introduce `this.observableProps` (downside: redudancy, ultimately confusion)
3) make `this.props` by default observable (downside: risk, subtle unforsee-able bugs, like having a prop that receives a plain object that must stay plain? use `asFlat`, might be confusing as well?)

andykog:
4) Make `@observer` configurable, something like:
`@observer({ inject: [], observableProps: true })`

mweststrate:
just took a shot at this one, by trying to make the `this.props` an observable reference. However that doesn't work out, because even if `shouldComponentUpdate` returns `false`, the nextProps props are still assigned to the component instance. So a straightforward `extendObservable(this, { props: asReference(this.props))` doesn't work out. Could be wrapping a getter / setter around it that only accepts new props if sCU or something, but it might become quite brittle. 

An alternative  `observableProps` property, and updating  that one in `componentWillUpdate` or `componentWillReceiveProps` might be easier, but sounds a bit confusing at the same time

mweststrate:
cool, just figured out how to do this [properly](https://github.com/mobxjs/mobx-react/commit/0bcf70bee3068a1b3df51b1969c28805625b59e5)!

[...]

PR #136, opened 2016-10-17
Observable props of observer-ed component

Strate:
see #124

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -97,26 +97,27 @@ const reactiveMixin = {
       || "<component>";
     const rootNodeID = this._reactInternalInstance && this._reactInternalInstance._rootNodeID;
 
+    let skipRender = false;
+
     function makePropertyObservableReference(propName) {
       let valueHolder = this[propName];
       const atom = new mobx.Atom("reactive " + propName);
-      const reportChanged = debounce(function reportChanged() {
-        atom.reportChanged();
-      })
       Object.defineProperty(this, propName, {
           configurable: true, enumerable: true,
           get: function() {
             atom.reportObserved();
             return valueHolder;
           },
-          set: mobx.action(function set(v) {
+          set: function set(v) {
             if (isObjectShallowModified(valueHolder, v)) {
               valueHolder = v;
-              reportChanged();
+              skipRender = true;
+              atom.reportChanged();
+              skipRender = false;
             } else {
               valueHolder = v;
             }
-          })
+          }
       })
     }
 
@@ -143,7 +144,9 @@ const reactiveMixin = {
             // If we are unmounted at this point, componentWillReact() had a side effect causing the component to unmounted
             // TODO: remove this check? Then react will properly warn about the fact that this should not happen? See #73
             // However, people also claim this migth happen during unit tests..
-            React.Component.prototype.forceUpdate.call(this)
+            if (!skipRender) {
+              React.Component.prototype.forceUpdate.call(this)
+            }
           }
         }
       });
--- a/src/observer.js
+++ b/src/observer.js
@@ -57,6 +57,33 @@ function patch(target, funcName) {
   }
 }
 
+function isObjectShallowModified(prev, next) {
+  if (null == prev || null == next || typeof prev !== "object" || typeof next !== "object") {
+    return prev !== next;
+  }
+  const keys = Object.keys(prev);
+  if (keys.length !== Object.keys(next).length) {
+    return true;
+  }
+  let key;
+  for (let i = keys.length - 1; i >= 0, key = keys[i]; i--) {
+    const newValue = next[key];
+    if (newValue !== prev[key]) {
+      return true;
+    } else if (newValue && typeof newValue === "object" && !mobx.isObservable(newValue)) {
+      /**
+       * If the newValue is still the same object, but that object is not observable,
+       * fallback to the default React behavior: update, because the object *might* have changed.
+       * If you need the non default behavior, just use the React pure render mixin, as that one
+       * will work fine with mobx as well, instead of the default implementation of
+       * observer.
+       */
+      return true;
+    }
+  }
+  return false;
+}
+
 /**
  * ReactiveMixin
  */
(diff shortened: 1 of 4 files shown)


Strate:
@mweststrate this PR contains bunch of fixes to reactive props/state. Latest version works perfectly with my current project (with strict mode enabled), so, from my point of view, it is good to merge.

mweststrate:
Thanks for finding and fixing this issue! Looking good. The automatic function conversion is a legacy concern indeed, I'll deprecate that in the next mobx version.

I was wondering whether we should actually do something with state. I think we can put that back to it's original check in SCU (`return this.state !== nextState`), since if you want observable state, you probably just use instance fields directly. Making state observable might actually break for people that rely on the async nature of `setState`.

I wanted to wrap this change up, and some related ones, once I am back from ReactiveConf (first week of november). Does that sound OK for you?

mweststrate:
Probably can be released earlier if you need it earlier as official build. Anyway, things left:

* [ ] thing about `state` observable or not
* [ ] docs
* [ ] release notes

Strate:
> Does that sound OK for you?

I am OK to use pre-release build for now, so, it is not need to release it asap. So, devote yourself for ReactiveConf :)

> Making state observable might actually break for people that rely on the async nature of setState

Does this PR breaks it? It defines setter for `state` prop, without any overloadings for `setState`.

>  if you want observable state, you probably just use instance fields directly

I am totally agree with you, but I want from library be as less surprise developer as possible. So, if developer refer to state from computed function, computation result should be invalidated on state changes (without this, developer will spent hours of debugging).

Strate:
@mweststrate seems that travis build is broken: I've just added failing test case, which fails `npm test` command on local machine, but travis build is green.

Strate:
@mweststrate I've changed the way how props/state observed.
I've reverted back your `shouldComponentUpdate` change, because if it always returns `false` and re-render triggers only with `forceUpdate` called by reaction on props/state modification, `componentWillUpdate`/`componentDidUpdate` callbacks called with same values on `nextProps`/`prevProps`.
I've made props/state observable via Atom class, and report about modification only if new props/state is not shallow equal with previous one.
But for now component rerenders twice: on standard React's update way and with reaction, because props are reactive.
Seems that there should be a check in reaction, if *only* props/state changed, do not force update component. I don't know how to do that.

mweststrate:
Ha though questions! I'm gonna take my time to think this through in two weeks :) But if you find a nice approach in the mean time let me know!

Strate:
I've already pushed the fix, done simply with boolean flag :) Just works, maybe there is a better solution :)

mweststrate:
Thanks a lot for your efforts! If you want btw I can make you part of the MobX organization, that allows you to push directly to (branches) of the mobx repo's. And it looks nice on your profile ;-)

Strate:
It would be nice;)

Strate:
@mweststrate could you please allow me write access to `mobx-react` npm package? I want to publish latest version of this PR (as a new version of `mobx-react@3.8.4-observableprops`). This is my npm account: https://www.npmjs.com/~eshenbrener

mweststrate:
Done

Op do 20 okt. 2016 om 10:05 schreef Artur Eshenbrener <
notifications@github.com>:

mweststrate:
Make sur eto use a `--tag` !

Op do 20 okt. 2016 om 10:18 schreef Michel Weststrate <mweststrate@gmail.com

Strate:
@mweststrate great, thanks. About your note for `--tag`. As for now there is 2 dist-tag on `mobx-react` package: `'dist-tags': { latest: '3.5.8', beta: '3.5.8-fix134' },`, there is no `observableprops` tag. I want to publish head of this PR under `3.8.5-observableprops.1` version, without `--tag` option (because there is no dist-tag for that version).
Is it OK, or I need to publish it under `beta` (or even `alpha`) tag?

mweststrate:
Hi @arthur, build that are not intended for general consumption should be
tagged with anything but "latest", so that they won't be installed
accidentally for everyone. So update the "version" field in package json to
"3.8.5-observableprops.1", then run "npm run build && npm publish --tag
beta".

Op do 20 okt. 2016 om 11:31 schreef Artur Eshenbrener <
notifications@github.com>:

Strate:
Published as `mobx-react@3.5.8-observableprops.1`

src/observer.js
```diff
@@ -3,6 +3,7 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import EventEmitter from './utils/EventEmitter';
 import inject from './inject';
+import debounce from "lodash.debounce";
```
Strate: Oh, no, I've forgot about this import. Gonna to remove.

mweststrate:
@Strate just testing this PR. Great work, thanks!

One question: Do we need to make state observable as well? After all `setState` is the only way currently to modify state in react, which will always trigger a rendering anyway? Although now `state` is optimized, where in the previous implementation any state change would trigger a re-rendering. Is that the intended change?

mweststrate:
Closed in favor of #154

Strate:
> Do we need to make state observable as well?

Yes, because developer could use state in computed as well as property.

> Although now state is optimized, where in the previous implementation any state change would trigger a re-rendering.

In a fact, state is not optimized from re-render point of view. As you can see, while `atom.reportChanged` called, `skipRender` set to true, this is not actually skip render of component, it is only prevent double-render. 1st caused by reactive props/state, second caused by standard react mechanics. So, `setState` always trigger re-render, this is not broken. We could add test case for this.

mweststrate:
@Strate what I thought indeed. could you the test case? after that feature should be good to go

mweststrate:
Preferably use the `v4` branch

alexhisen:
This is a potentially breaking change. I've submitted a PR with some doc updates that reflect this change but it may need to be more prominent. Here is a use case where this is a breaking change: https://alexhisen.gitbooks.io/mobx-recipes/content/observable-based-routing.html

PR #153, opened 2016-11-07
Fixed propType usage typo in readme

loktar00:


mweststrate:
actually the example is correct (it is consistent with React), but both are experted. The typings however only included `propTypes`, I'll update that. Thanks for the PR nonetheless!

PR #169, opened 2016-11-23
Fix order of (recently added) type definitions from most to least specific

timmolendijk:
Just found out that the order of the type definitions I added (v4.0.1) is not right. Something to do with specificity as far as I can see. Didn't work as intended before. Does work with this PR.

(I hadn't found out earlier because I had tested them as a merged declaration, in which new definitions are _prepended_ the existing ones I suppose.)

Sorry!

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -17,10 +17,10 @@ export function observer<P>(stores: string[], clazz: React.ComponentClass<P>): R
 export function observer<P>(stores: string[]): <TFunction extends React.ComponentClass<P>>(target: TFunction) => TFunction; // decorator signature
 
 // inject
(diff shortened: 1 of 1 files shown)


mweststrate:
Woops, merged to soon, test fail now. With which typescript version were you testing? The typescript tests fail with this change (see `tsc -p test/ts --pretty`)

mweststrate:
anyway, I'll look into it, needs some clean up anyway

mweststrate:
@timmolendijk could you check whether everything still works correctly with `mobx-react@4.0.2-rc` ?

timmolendijk:
All good now!

Issue #174, opened 2016-12-01
Type definition for `inject` does not support outer props signature being different from inner props signature


timmolendijk:
This will work:
```tsx
inject((stores: any, props: { myProp }): { myProp } => props)(({ myProp }) => <strong>hurray</strong>);
```

This will not work:
```tsx
inject((stores: any, props: { myProp }): { anotherProp } => ({ anotherProp: 'whoops' }))(({ anotherProp }) => <strong>{anotherProp}</strong>)
```

The culprit is the type definition of `inject` assuming that the signature of the wrapped component is the same as the signature of the created (wrapping) component.

mweststrate:
Closed, seems not possible to express this decently in TS

PR #175, opened 2016-12-01
Allow signature of injecting component to differ from signature of injected component

timmolendijk:
Address #174


timmolendijk:
Hmm, failing tests. Am I missing anything?

tkrotoff:
You should write a test case here: https://github.com/mobxjs/mobx-react/blob/master/test/ts/compile-ts.tsx and verify everything works with `npm run test`. This way you ensure there will be no regression in the future.

timmolendijk:
FYI: I am no longer using this package, so I won't find time to look into this any time soon.

mweststrate:
Thanks for the update, closing for now.

Issue #182, opened 2016-12-20
Sub-Provider store as computed value throws Warning


dropfen:
Our structure looks like this:

Very small pseudo example

```
<Provider storeOne={storeOne}>
  <Router>
    <Route 
       path="/some/:id"
       onEnter={(paramsId) => { storeOne.setId(paramsId)} } component={Component} 
    />
  </Router>
</Provider>
```

storeOne has a computed value, based on the current Id

```
class storeOne {
   @observable id;
   items = [];
   @action setId(id) => this.id = id;
   @computed get current() => this.items[this.id];
}
```

in Component render we use the relative item of the store which is a computed value

```
<Provider item={storeOne.current} />
   <ItemConsumerOne />
   <ItemConsumerTwo />
   <ItemConsumerThree />
<Provider>
```

The problem now is, that every time we change the url, we see that warning:

https://github.com/mobxjs/mobx-react/blob/master/src/Provider.js#L41

What's the point of it? How could we suppress it, or do you see the critical points, we are not aware of? thank you!

mweststrate:
The reason for the warning is that replacing a store is a change that might not make it to all descendents, if there is any `shouldComponentUpdate` in between that update might never arrive. (see https://medium.com/@mweststrate/how-to-safely-use-react-context-b7e343eff076L for the reasons why). This could theoretically solved in mobx-react, but I think the proper way to think about `Provider` is as _dependency injection_, which as a thing that happens once. 

A simple workaround is probably to inject your store instead of item (possibly by using `@inject(stores => ({ item: stores.storeOne.current }))` if you don't want to be aware of `storeOne` in your component itself

dropfen:
Thank you for the fast reply, it makes sense. However in our scenario we are aware what we are doing, some of us said we should suppress the warning explicitly if possible ;)

dropfen:
Would you accept a pr like `if props.suppressChangedStoreWarning => no warning` ?

mweststrate:
yeah that sounds reasonable :)

dropfen:
is coming ;)

dropfen:
@mweststrate the Link in your comment did't work

mweststrate:
Thanks! Released as 4.0.4

dropfen:
Great, thank you!

alisabzevari:
What is the best practice of initializing a store based on another one and put both of them in `Provider`?
[...]

PR #183, opened 2016-12-20
Add warning suppression for changed stores, and a test for it

dropfen:
fixes: #182


jzaefferer:
Probably also needs docs updates, maybe mention the new prop somewhere under https://github.com/mobxjs/mobx-react#provider-and-inject ?

dropfen:
Note added.

dropfen:
@jzaefferer You are right, I added it extra because it makes it clear, that this is a boolean. should't be a blocker.

mweststrate:
Looking good, thanks!

PR #188, opened 2016-12-26
Quick Grammar fix to README.md

benmccormick:
.  Just a quick fix. Now "decorators are **documented** here", rather than "decorators are **document** here.


mattruby:
Thanks!

Issue #97, opened 2016-08-03
Observer decorator in TSX throws type error when updating project from 3.3.0 to 3.5.x


Haubach:
I'm working in a code base which has a react class decorated with the observer decorator. In Mobx-react 3.3.0 this worked perfectly, but after installing 3.5.x it throws a type error on compilation with typescript 1.8.10.

> Unable to resolve signature of class decorator when called as an expression.
  Type 'ComponentClass<void>' is not assignable to type 'void'.

[Here's](https://github.com/Haubach/Mobx-react-bug) a minimal example which recreates the issue.

`@observer
class Foo extends React.Component<void,void> {
	render() {
		return (<div></div>);
	}
}`

mweststrate:
Quick check, what about `React.Component<{}, {}>` (that's what I always use)?

Haubach:
Yes, that does work.

andykog:
Strange. Component state should allow `void` https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/react/react.d.ts#L16
@Haubach, are you using that type definition for React? Works fine for me.

Haubach:
Yes, that is the react type definition I'm using in [the example repo](https://github.com/Haubach/Mobx-react-bug).

I've just verified the behavior on my colleagues machine.

mweststrate:
It is unclear to me why `void` isn't allowed by the type system, could be investigated further but is `{}` instead of `void` an acceptable work around? (Personally I would argue that it is more correct as well, props aren't absent, but just empty)

Haubach:
The decision to use void over empty object was taken some time ago where we encountered an issue in another code base, but I don't remember the specific reasoning. For now, this regression is not a serious issue and hopefully it will continue to be a non-issue if you don't intend to continue to support the 3.3.0 behavior.

PR #193, opened 2017-01-11
Allow Component<void, void>

tkrotoff:
Fix https://github.com/mobxjs/mobx/issues/665 and #97

Now you can write:
```TypeScript
@observer
class ComponentWithoutPropsAndState extends Component<void, void> {
    componentDidUpdate() {
    }

    render() {
        return (<div>Hello!</div>)
    }
}
```

The solution was a simple `| void` inside index.d.ts


mweststrate:
Thanks!

PR #205, opened 2017-02-16
Fix small typos in README.MD

vkrol:


README.md
```diff
@@ -358,7 +358,7 @@ const mountedComponent = mount(
 ```
 
 Bear in mind that using shallow rendering won't provide any useful results when testing injected components; only the injector will be rendered.
-To test with shallow rendering, instantiate the `.wrappedComponent instead:`: `shallow(<Person.wrappedComponent />)`
+To test with shallow rendering, instantiate the wrapped сomponent instead: `shallow(<Person.wrappedComponent />)`
```
mattruby: I think this should be: 'wrappedComponent' instead:

mattruby:
Thanks for the fix!

PR #211, opened 2017-03-05
remove shouldComponentUpdate check

aihornmac:
I was using `react-router@4.0.0-beta.7` with `mobx-react/observer`.

`react-router@4` uses `Router#setState` to trigger re-render from history changes, which means every component that specifies `shouldComponentUpdate` must be responsible for location changes.

One solution is to use `@withRouter` along with `@observer`, which might be really redundant.

Given the fact that current code of `shouldComponentUpdate` defined in `observer` is behaving like `PureRenderMixin`, I suggest removing it since if users really need it they would have defined it or used `PureRenderMixin`/`React.PureComponent`, it shouldn't harm much restoring to default behavior.

Looking forward to your response. Thank you for your dedication!

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -224,19 +224,12 @@ const reactiveMixin = {
     }
   },
 
-  shouldComponentUpdate: function(nextProps, nextState) {
+  shouldComponentUpdate: function() {
     if (isUsingStaticRendering) {
       console.warn("[mobx-react] It seems that a re-rendering of a React component is triggered while in static (server-side) mode. Please make sure components are rendered only once server-side.");
     }
-    // update on any state changes (as is the default)
-    if (this.state !== nextState) {
-      return true;
-    }
-    // update if props are shallowly not equal, inspired by PureRenderMixin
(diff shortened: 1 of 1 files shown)


mweststrate:
The point of `@observer` / mobx-react is that `sCU` doesn't need to be implemented by users and one still gets optimized performance out of the box. If you can setup a test project or testcases in the testsuite, it can be investigated if the check can be made smarter, for example skipping the hook for `setState` calls etc.

aihornmac:
Yes, I understand that mobx is doing partial update. But still there is such case where default behavior is wanted, and since react-router is not bound to mobx, this feature is probably needed.

mweststrate:
Closing this issue, the impact of people needing to use `@withRouter` for react-router related components is a lot smaller than people suddenly having to provide a shouldComponentUpdate strategies, which is one of the biggests pro's of mobx-react

mweststrate:
Note that it is still possible to override `shouldComponentUpdate` in specific components if you like to

PR #222, opened 2017-03-17
Add react-native field to package.json

farwayer:
With this change you can import from `native.js` by
```js
import {observe} from 'mobx-react'
``` 
React Native packager will automatically use correct file.


mweststrate:
Great! Thanks

PR #240, opened 2017-04-10
Webpack 2

andykog:


andykog:
Travis seems to work fine after upgrading node version.

Issue #221, opened 2017-03-16
Exceptions in render don't get passed into `extras.onReactionError`


le0nik:
This issue came up after this change https://github.com/mobxjs/mobx-react/commit/456bf475c33b3fdd003dfdb7d8871d1e58e4edcf

Or is there another way to attach logging behavior for these exceptions?

mweststrate:
Correct, the exceptions don't escape anymore indeed. Where do you use this for? Logging exceptions in prod? Would a global onError hook for mobx-react specifically help?

le0nik:
I send exceptions like this to `sentry` in production. Yes, global `onError` in `mobx-react` would definitely help! It's a bit inconsistent in my opinion, though. Docs for `extras.onReactionError` state that all errors inside reactions get there, but it's not the case for errors in `mobx-react` render. Instead they bubble up to the global scope.

mweststrate:
global `onError` in mobx-react package is now available in 4.2.1

PR #262, opened 2017-05-28
Global errors handler

andykog:
See #221

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -17,6 +17,9 @@ export function observer<P, TFunction extends React.ComponentClass<P | void>>(ta
 export function inject<P>(...stores: string[]): (<TFunction extends IReactComponent<P>>(target: TFunction) => (TFunction & IWrappedComponent<P>)); // decorator signature
 export function inject<T, P>(storesToProps : IStoresToProps<T, P>): (<TFunction extends IReactComponent<T | P>>(target: TFunction) => (TFunction & IWrappedComponent<T>)); // decorator
 
+export function onError(cb: () => void): void;
+export function offError(cb: () => void): void;
+
 // Deprecated: observer with with stores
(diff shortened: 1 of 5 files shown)


src/observer.js
```diff
@@ -205,8 +211,10 @@ const reactiveMixin = {
           this.__$mobRenderEnd = Date.now();
         }
       });
-      if (exception)
+      if (exception) {
+        errorsReporter.emit(exception);
```
mweststrate: If there is an error reporter, should we still throw? (I think so)

src/index.d.ts
```diff
@@ -17,6 +17,9 @@ export function observer<P, TFunction extends React.ComponentClass<P | void>>(ta
 export function inject<P>(...stores: string[]): (<TFunction extends IReactComponent<P>>(target: TFunction) => (TFunction & IWrappedComponent<P>)); // decorator signature
 export function inject<T, P>(storesToProps : IStoresToProps<T, P>): (<TFunction extends IReactComponent<T | P>>(target: TFunction) => (TFunction & IWrappedComponent<T>)); // decorator
 
+export function onError(cb: () => void): void;
```
mweststrate: this is a little inconsistent with most mobx api's, where event handler registration returns a disposer functions, instead of having a separate off...

src/observer.js
mweststrate: On the other hand, the throw doesn't add that much here, since Reactions catch all exception, the only additional value of this throw is that the error get's printed to the console

andykog: > only additional value of this throw is that the error get's printed to the console
I think, that's a big deal, as `onError` is meant to be used only in specific scenarios (tracking errors with sentry etc.)

src/index.d.ts
andykog: You are right, fixed

src/observer.js
mweststrate: Check, let's leave it this way

PR #263, opened 2017-05-28
add missing typings for PropTypes

panjiesw:
add missing `observableArrayOf` and `arrayOrObservableArrayOf` of PropTypes typing.


andykog:
Thanks!

PR #265, opened 2017-05-29
fix rollup umd runtime error re: module is not defined

rossipedia:
* ES2015 doesn't define `module`, nor `module.exports`
* Rollup's UMD wrapper doesn't pass in `module`, but it does pass in `exports`
* Since a bare variable name in JS is valid, and assumed to reference a variable in ambient/global scope, we can just use `exports` instead of `module.exports`


mweststrate:
Confirmed fix, released as 4.2.1

PR #282, opened 2017-06-19
I add some magic to inject 

sheweichun:
Before the modification,we use inject like beblow:
`
@inject("color")
`
It is ok,But if I want to only visit deeper attributes like a.b,I can only do like this:
`
@inject("a")
`
then,visit b by this.props.a.b

after the modification,we can do like this:
`
@inject("a.b:propName")
`
then,visit b by this.props.propName

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -72,15 +72,47 @@ function createStoreInjector(grabStoresFn, component, injectNames) {
   return Injector;
 }
 
+/**
+ * Parse injection name
+ */
+function parseStoreName(storeName){
+  let storeNameSplited = storeName.split(':');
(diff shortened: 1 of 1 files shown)


mweststrate:
This is the kind of logic that imho doesn't belong to this package. First of all it needs a lot of documentation, and there are many libraries that do this properly better. Like a lodash pluck.  Furthermore this problem is easy to be solved in userland. You can simply create your own HoC by leveraging the fact that inject can also take functions, so you could do something like:

```
function pluckInject(path) {
   return inject(stores => pluck(stores, path))
}

@pluckInject("a.b.c") class MyComponent....
```

Issue #270, opened 2017-06-05
Reorder observer declarations for better typescript error messages


gbishop:
I have an idea. Typescript produces an incomprehensible error message when observer is applied to classes or functions that are slightly wrong. Here is a small example:

```
import * as React from 'react';
import { observer } from 'mobx-react';

@observer
class Comp extends React.Component<void, null> {
  render() {
    return (
      <p>Hello</p>
    )
  }
}
```
The error message produced by tsc is:
```
src/test.tsx(4,2): error TS2345: Argument of type 'typeof Comp' is not assignable to parameter
 of type 'string[]'.
  Property 'push' is missing in type 'typeof Comp'.
```
As a newb to mobx I can tell you that the reference to string[] completely blew my mind!

A simple change to index.d.ts produces this much more useful error message:
```
src/test.tsx(4,2): error TS2345: Argument of type 'typeof Comp' is not assignable to parameter of type 'ComponentClass<void | {}>'.
  Type 'Comp' is not assignable to type 'Component<void | {}, ComponentState>'.
    Types of property 'setState' are incompatible.
      Type '{ <K extends never>(f: (prevState: null, props: void) => Pick<null, K>, callback?: (() => any) | ...' is not assignable to type '{ <K extends never>(f: (prevState: ComponentState, props: void | {}) => Pick<ComponentState, K>, ...'.
        Types of parameters 'f' and 'f' are incompatible.
          Type '(prevState: ComponentState, props: void | {}) => Pick<ComponentState, any>' is not assignable to type '(prevState: null, props: void) => Pick<null, any>'.
            Types of parameters 'prevState' and 'prevState' are incompatible.
              Type 'null' is not assignable to type 'ComponentState'.
```
Now the error relates to what really is going wrong. The last bit tells me that I shouldn't have used null in that second position. Similarly confusing messages are produced when functional components may return null. 

This origin of the confusing message is that typescript appears to use the last declaration to produce the error message.

My suggestion is to simply reorder the declarations so that the deprecated versions come first. That is, move lines 22-26 in index.d.ts to above line 13.

mweststrate:
The current typings are afaik in the current order to get all scenarios working. Feel free to open a PR to reorder them if it improves error messages, just make sure all the current typescript tests keeps succeeding: https://github.com/mobxjs/mobx-react/blob/master/test/ts/compile-ts.tsx

gbishop:
I will attempt it next week when I return to my office. I'll report back here.

[...]

PR #290, opened 2017-07-01
Move deprecated observer defines in front of current.

gbishop:
This improves error messages because tsc refers to the last it sees.

npm run test:ts completes without complaint. Issue #270.


mweststrate:
Thanks!

PR #293, opened 2017-07-05
Fixed findDOMNode if-condition when used with non-DOM renderers

normano64:
When enabling `trackComponents`, `findDOMNode` was throwing exceptions in the `try...catch` statement when used with non-DOM renderers.
I guess the empty-module behaviour changed when moving from webpack to rollup?

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -4,6 +4,7 @@ import ReactDOM from 'react-dom';
 import EventEmitter from './utils/EventEmitter';
 import * as PropTypes from './propTypes';
 import inject from './inject';
+import empty from '../empty';
 
 /**
  * dev tool support
@@ -19,7 +20,7 @@ export const componentByNodeRegistery = typeof WeakMap !== "undefined" ? new Wea
 export const renderReporter = new EventEmitter();
 
 function findDOMNode(component) {
-  if (ReactDOM) {
+  if (ReactDOM !== empty) {
     try {
       return ReactDOM.findDOMNode(component);
     } catch (e) {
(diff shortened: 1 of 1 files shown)


src/observer.js
```diff
@@ -19,7 +20,7 @@ export const componentByNodeRegistery = typeof WeakMap !== "undefined" ? new Wea
 export const renderReporter = new EventEmitter();
 
 function findDOMNode(component) {
-  if (ReactDOM) {
+  if (ReactDOM !== empty) {
```
mweststrate: if think `ReactDOM && ReactDOM.findDOMNode` would be a lot clearer to read?

mweststrate:
Are you importing from `mobx-react` with custom renderers? Instead you probably should import from `mobx-react/custom`, which strips away the whole dev tool integration which doesn't have much value without DOM rendering.

normano64:
Hi @mweststrate, currently the if-case isn't stripped from `custom.js`.

Here's snippet from `custom.js`:
```js
var empty = {};
/* ... */
function findDOMNode(component) {
  if (empty) {
    try {
      return empty.findDOMNode(component);
    } catch (e) {
      // findDOMNode will throw in react-test-renderer, see:
      // See https://github.com/mobxjs/mobx-react/issues/216
      // Is there a better heuristic?
      return null;
    }
  }
  return null;
}
```
The if-case is still remain when using either `ReactDOM !== empty` or `ReactDOM && ReactDOM.findDOMNode`.

src/observer.js
normano64: Good point. 👍

mweststrate:
Merged, thanks!

normano64:
Thanks! Sorry for the delay.

PR #305, opened 2017-08-02
Add reference to component to reaction

### LINKED ISSUE #279: Reference Component in Reaction
It would be very handy for debugging tools if the reaction created [here](https://github.com/mobxjs/mobx-react/blob/4177daa6685dc6c40a232aeb297f93ee27675bde/src/observer.js#L166) has an explicit reference to the component that is being rendered.
Something like `reaction.react_component = this;` would do it.

hawkins:
Fixes #279

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -189,6 +189,7 @@ const reactiveMixin = {
           }
         }
       });
+      reaction.reactComponent = this;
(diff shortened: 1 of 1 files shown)


mweststrate:
Merged, thanks!

PR #321, opened 2017-09-08
Introduce prettier

realityforge:
Introduce prettier to project using the same setting as was applied to the main mobx repository by https://github.com/mobxjs/mobx/pull/1103


mweststrate:
Thanks!

mweststrate:
Merged manually, closing

PR #340, opened 2017-10-08
Add `displayName` to Observer component

nickbreaton:
Hello,

The Observer component renders improperly in React dev tools:

| Current | After Fix |
| -----|----|
| image | image | 

The fix was a super simple one-liner.

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -340,6 +340,8 @@ function mixinLifecycleEvents(target) {
 // TODO: support injection somehow as well?
 export const Observer = observer(({ children }) => children())
 
+Observer.displayName = "Observer"
+
 Observer.propTypes = {
     children: (propValue, key, componentName, location, propFullName) => {
         if (typeof propValue[key] !== "function")


tkrotoff:
Do you use TypeScript? Do you use a React SFC?
See https://github.com/facebook/react-devtools/issues/267#issuecomment-333099574

This is what I do in my code:
```TypeScript
const Example = observer(({...}) =>
  <div>Example</div>
);
Example.displayName = 'Example';
```

Not sure introducing `Observer` is a good idea.
- User finds `Observer` name within React DevTools
- He tries to find an Observer component in his code and can't find it

On the other hand, if React DevTools displays `_class` the dev knows something is wrong.

At least `Observer` should be renamed to something like `MobxObserver`.

nickbreaton:
Hello @tkrotoff,

I'm not sure what you mean by "introducing `Observer`". It is already part of the API.

https://github.com/mobxjs/mobx-react/blob/master/src/observer.js#L341

I am simply adding the `displayName` so it shows properly in the DevTools. The code that produces the above component tree is:

```jsx
import { Observer } from 'mobx-react'

const App = () => (
  <Observer>
    {() => (
      <h1>MobX is Awesome!</h1>    
    )}
  </Observer>
)

ReactDOM.render(<App />, ...)
```

tkrotoff:
> I'm not sure what you mean by "introducing Observer"

Talking about the display name. Not the React component but the string you've added in you PR: `Observer.displayName = "Observer"`

nickbreaton:
I'm just not sure why we wouldn't want the name of the component in the code, to match what is shown in the DevTools. If a dev imports `Observer` from mobx-react, then shouldn't it be reflected in the tools?

The reason this is not being done by default, is because the stateless component that is being wrapped by `observer()` is anonymous.

React Motion has a similar API, (function as a child), and you can see in the DevTools on their example site, their component is named `Motion`.

image

[Link to Demo](http://chenglou.github.io/react-motion/demos/demo0-simple-transition/)

Also, to answer your question about TypeScript. Yes I was using TypeScript, but mobx-react should not even be running through TypeScript as it is already compiled to ES5.

Please let me know if I am missing something.

tkrotoff:
> React Motion has a similar API

With React Motion you write [`<Motion>...</Motion>`](https://github.com/chenglou/react-motion/search?utf8=%E2%9C%93&q=motion&type=), right? (I don't use React Motion so not sure). So it's not the same as MobX `observer(...)` which is a decorator/function that wraps the real React Component.

nickbreaton:
I've thrown together a CodeSandbox to hopefully make everything more clear. Sorry for the delay. 

It can be found [here](https://codesandbox.io/s/k96olm09xo), with a direct link to the application [here](https://k96olm09xo.codesandbox.io/) (easier to look at React DevTools)

You say:

> there is no `Observer` component in your code

however, that is not true. Just like in react-motion you `import { Motion } from 'react-motion'`, you can `import { Observer } from 'mobx-react'`.  

You can see in the provided CodeSandbox that this is the case. I understand that in the implementation `observer()` is being used in the implementation of the `Observer` component, however this should not matter to the consumer of the mobx-react API. For all I know react-motion could be using a HOC under the hood as well, but it does not and should not matter to me.

To be totally clear, if you check the React DevTools on the [CodeSandbox App](https://k96olm09xo.codesandbox.io/) you can see that `<Motion>` shows up, but `<Observer>` does not. In the place of where `<Observer>` should be, lies `<_class>`, which gives no context as to what that component is being used for.

mweststrate:
@tkrotoff I think the confusion in this discussion is `observer(Component)` versus `Component = <Observer>{() => stuff}</Observer>`. The latter indeed generates a different component tree with a HoC `Observer` (currently unnamed). The latter is what @nickbreaton is attempting to fix, which seems to be correct.

Merging now, but if I overlooked something let me know @tkrotoff :)

tkrotoff:
You are right,
now `<Observer>{() => stuff}</Observer>` displays `Observer` in React DevTools and it does not affect `observer()` => that was my worry.

You still need and should write:
```TypeScript
const Example = observer(({...}) =>
  <div>Example</div>
);
Example.displayName = 'Example';
```

PR #343, opened 2017-10-10
Remove void from React.ComponentClass<P | void>

tkrotoff:
Not needed anymore thx to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/17288 (and btw it was found that `{}` is better than `void` for props).
`void` was introduced by me in #193


mweststrate:
Thanks!

PR #350, opened 2017-10-31
Edit README to add another strong typing suggestion

hayeah:


mweststrate:
I think this suggestion is too ugly to propose as 'solution'. Even `// @ts-ignore` would be neater ;-)

hayeah:
hmm, it seems that `@ts-ignore` only works for `.js`?

The recommendation for `.ts` is to cast to `any`. https://github.com/Microsoft/TypeScript/issues/9448

The suggestion in the current README is suitable if strict null check is not enabled. With null-checking enabled nullable injected store is quite painful to use.

mweststrate:
Note that the `!` operator makes it a lot more bearable with strict null
mode `this.props.store!.property`

Op do 2 nov. 2017 om 14:20 schreef Howard Yeh <notifications@github.com>:

mweststrate:
Or store the inject store once on the component instance, in constructor:
`this.myStore = this.props.myStore!`

Op do 2 nov. 2017 om 14:21 schreef Michel Weststrate <mweststrate@gmail.com

hayeah:
oh I didn't know about the `!` operator! Thanks, that's much better.

PR #368, opened 2017-11-27
Fix debug name for react fiber

andykog:
Discovered in https://github.com/andykog/mobx-devtools/issues/13


andykog:
Or, maybe, remove those usages of React private API's?

mweststrate:
Thanks! Yeah maybe we should remove it in the future if it causes anymore trouble.

PR #376, opened 2017-12-11
Update Observer component definition

RafalFilipek:
Hi!

Right now we have to wrap observerd value with HTML tag like:

```
const C => <Observer>{() => <div>{obser.vable}</div>}</Observer>
```

With this change you can just return observed value

```
const C => <Observer>{() => obser.vable}</Observer>
```

Thank you!

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -65,7 +65,7 @@ export function onError(cb: (error: Error) => void): () => void
 
 export class Provider extends React.Component<any, {}> {}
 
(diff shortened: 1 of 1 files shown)


mweststrate:
Great, thanks!

FredyC:
@mweststrate Do you plan to release a new version soon? This one is particularly annoying since I cannot return `null` from there.

mweststrate:
@FredyC yeah, I want to release if #403 is in. If that is not today, I'll probably cut a release at the end of the day anyway

PR #402, opened 2018-01-27
Make `isObjectShallowModified` treat `NaN` equal to `NaN`

### LINKED ISSUE #363: NaN in props causes observer not to rerender in React 16
https://github.com/mobxjs/mobx-react/blob/bc10762c46726831da9b8ae9f7180a3aea1ccd15/src/observer.js#L140-L142

The logic that sets `skipRender` to true, skipping a `forceUpdate`, checks to see if the current props and next props are shallow equal to each other with `isObjectShallowModified`. The business logic as I understand it is basically “if props are shallow modified, don’t call `forceUpdate` because React will update the component on its own.”

However, this assumption appears to be wrong as of React 16 when the value of one of those props is `NaN`, because while `NaN !== NaN`, apparently React’s own handling of `NaN` has changed, and React does _not_ in fact update the component. The result is that if you pass `NaN` as a prop to any observer component, it will never update as a reaction to observables changing. E.g.

```jsx
@observer
class Foo extends Component {
  render() {
    return /* something that hits any number of observables */;
  }
}

// elsewhere
<Foo bar={NaN} /> // This instance of Foo will never update
```

Live example: https://www.webpackbin.com/bins/-KzQSm3N0aWfc_U2V1BZ

---

I'm happy to fix this myself, but need some guidance. If we make `isObjectShallowModified` compare values of `NaN` as “not modified,” this will cause double rerenders in React 15 and below. Is that an acceptable tradeoff, or should we detect which version of React is being used?

andrewbranch:
Sorry for the delay!

Fixes #363

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -94,6 +94,9 @@ function isObjectShallowModified(prev, next) {
     }
     let key
     for (let i = keys.length - 1; i >= 0, (key = keys[i]); i--) {
+        if (isNaN(next[key]) && isNaN(prev[key])) {
+            continue
+        }
         if (next[key] !== prev[key]) {
             return true
         }
--- a/test/observer.test.js
+++ b/test/observer.test.js
@@ -91,6 +91,10 @@ describe("nestedRendering",async()=>{
         expect(getDNode(store.todos[0], "title").observers.length).toBe(1)
     })
 
+    test("should rerender even when passed a prop with value `NaN`", ()=>{
+
+    })
+
     test("rerendering with outer store added",()=>{
         store.todos.push({
             title: "b",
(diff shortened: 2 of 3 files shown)


mweststrate:
Thanks for the PR. 

Note that some of the tests where failing in this branch, because the `isNaN` check was not correct (cause `isNaN({}) === true`). 

Next time make sure all tests pass :)

Merged now manually

mweststrate:
I just added my VSCode debugging settings to the repo. That might make it
easier for next time. Beyond that I use mainly `yarn jest --watch`

Op do 1 feb. 2018 om 20:00 schreef Andrew Branch <notifications@github.com>:

Issue #387, opened 2018-01-09
Observer with render prop and inject


FredyC:
I would like to propose an alternative way of using `Observer`. There is a clear benefit, that you don't need to repeat closing element there and it feels bit cleaner in my opinion.

```js
<Observer
  render={() => (
    <div />
  )}
/>
// vs (note: both are formatted by Prettier)
<Observer>
  {() => (
    <div />
  )}
</Observer>
```

Furthermore, it's a bit of pain needing to use `inject` HOC with this pattern. It separates two concerns that should stick together in my opinion. What about something like this?

```js
<Observer inject={[ 'store1', 'store2' ]} render={({ store1, store2 }) => (
  <div>{store1.observable && store2.observable}</div>
)} />
```

I am not entirely sure it's possible to achieve this with the current implementation of the `Observer` which is just wrapped HOC under the hood. Perhaps the whole thing could go other way around, implement `Observer` as the regular class with everything and then turning it to HOC is much easier.

mweststrate:
@FredyC that looks nice indeed. I think something like that could quite easily be implemented as just sugar around the current api's. Feel free to do so!

mweststrate:
Released in 4.4.0

urugator:
I think that introducion of `inject` prop on `Observer` wasn't such a good idea as it can make `props` handling quite tricky:
```javascript
const MyCmp = props => <Observer 
  inject={(stores, props) => ({ store: stores.store, prop: props.prop })}
  render={props => (
    <div>
      props.store.something + props.prop      
    </div> 
  )}
/>
```
There are 3 different props objects:
Props from surrounding context (in this case props accepted by `MyCmp`)
Props passed to `inject` mapper, which are always empty...
Props passed to `render` which are the props returned from `inject` mapper...

Basically the problem is that `Observer` doesn't accept any props on it's own, therefore `inject` mapper has no props to intercept. Therefore I think the injection should be limited to "static" injection or removed altogether.

See #419

urugator:
@FredyC Seems it tricked you as well ... have you noticed that `prop` is always `undefined`?

FredyC:
Well, that's a different problem that `inject` attribute is not exposing props in the second argument and I don't even think it's needed. This role is well suited by eg. `withProps` from `recompose` library. I wouldn't be dragging it here. Yea I know that's HoC world so it's kinda nicer in here. But if anything, I would introduce different prop for mapping component props then mixing these two concerns together.

[...]

PR #403, opened 2018-01-30
Add Observer render and inject sugar #387

Sunshine168:


FredyC:
Just a hint, it might be easier to review this if you turn off Prettier :) (or perhaps eslint autofix?). Any kind of reformatting should be done in a separate PR in my opinion.

Sunshine168:
Oh  i got a hobby which formate code anytime. i check package seem  `husky`  not be added, so precommit did not work well. 
i thought the rep should be format by same coding style and precommit should work, following this the code will be easier to review

hhhh,  i very appreciate your suggestion ,so i did it and will update typescript definitions later

FredyC:
Yea, you are right, husky is not installed so `precommit` hook is not working. I've made another PR #407 to fix that. Once that get merged, you can simply rebase this and only relevant changes will be part of this PR.

mweststrate:
@Sunshine168 Thanks for the PR! @FredyC's PR is merged, could you rebase?

PR looks good :). One remaining question; could you update the Readme as well describing this alternative syntax?

FredyC:
Also, it's obvious that using `children` has a priority, but there should be test showing a use of both at once. And reflect this in docs as well.

Sunshine168:
i will rebase later  , i am still in working

Sunshine168:
done !

README.md
```diff
@@ -82,7 +82,7 @@ It takes as children a single, argumentless function which should return exactly
 The rendering in the function will be tracked and automatically re-rendered when needed.
 This can come in handy when needing to pass render function to external components (for example the React Native listview), or if you
 dislike the `observer` decorator / function.
-
+And it can work with `Provider` use inject  and render property / or just takes as children, detail in Example2.(PS:using children has a priority than render , so dont use at the same time)
```
FredyC: I would split this, there are two concerns to cover - render prop and injecting. Also, I have a trouble understanding what it says. Let me rephrase this.

> In case you are a fan of render props, you can use that instead of children. Be advised, that you cannot use both approaches at once, children have a precedence.

_Here goes example showing render prop only_

> Observer can also inject the stores simply by passing a selector function.

_Second example with inject_

README.md
```diff
@@ -105,6 +105,38 @@ React.render(<App person={person} />, document.body)
 person.name = "Mike" // will cause the Observer region to re-render
 ```
 
+Example2:
+
+```javascript
+class App extends React.Component {
+  render() {
+     return (
+          <Provider h="hello" w="world">
```
FredyC: Are you sure you can pass a string as a store prop? I am not aware that it would turn to observable automagically here. In that case, it's kinda misleading example.

src/observer.js
```diff
@@ -349,12 +349,41 @@ function mixinLifecycleEvents(target) {
...
+        return component()
+    }
+    const InjectComponent = inject(observerInject)(component)
+    return <InjectComponent />
+})
 
 Observer.displayName = "Observer"
 
 Observer.propTypes = {
+    render: (propValue, key, componentName, location, propFullName) => {
```
FredyC: DRYing this would be nice :)

README.md
FredyC: Oh and with injecting it's worth to show an example that passing strings work same way as with `inject` HoC.

Sunshine168:
@FredyC  Ohhhh  very thank you for your help    .tat   now will be better?

src/observer.js
Sunshine168: ennnn Any suggestion to diy  Observer.propTypes

README.md
```diff
@@ -105,38 +103,63 @@ React.render(<App person={person} />, document.body)
...
-     )
-  }
-}
+
 const Comp = () => (
     <div>
         <Observer
-            inject={store => ({ h: store.h, w: store.w })}
-            render={props => <span>{`${props.h} ${props.w}`}</span>}
+            render={() => <span>hello world</span>}
```
FredyC: Hm, not very good example since there is nothing to observe :D Perhaps pass some variable in props to Comp and use that in render so it makes more sense.

README.md
```diff
@@ -105,38 +103,63 @@ React.render(<App person={person} />, document.body)
 person.name = "Mike" // will cause the Observer region to re-render
 ```
 
-Example2:
+In case you are a fan of render props, you can use that instead of children. Be advised, that you cannot use both approaches at once, children have a precedence.
+
+Here goes example showing render prop only
```
FredyC: Well, I did not mean this line to be included there as well, that's why I used italic ;)

src/observer.js
FredyC: Well, if you look closely, it's the same function except for that first condition. Which doesn't make any sense to me. It means, that when checking `render` prop, you are telling it's ok if `children` prop is a function. That doesn't make any sense to me.

README.md
```diff
@@ -105,6 +103,56 @@ React.render(<App person={person} />, document.body)
 person.name = "Mike" // will cause the Observer region to re-render
 ```
 
+In case you are a fan of render props, you can use that instead of children. Be advised, that you cannot use both approaches at once, children have a precedence.
+
+Here goes example showing render prop only
```
FredyC: This line is a bogus one, I did not mean to include it by using _italic text_ :)

README.md
```diff
@@ -105,6 +103,56 @@ React.render(<App person={person} />, document.body)
...
+
+const person = observable({ name: "John" })
+
+React.render(<App person={person} />, document.body)
+person.name = "Mike" // will cause the Observer region to re-render
+```
+
+Observer can also inject the stores simply by passing a selector function.
+
+Example with inject
```
FredyC: Same for this line, it's redundant imo.

src/observer.js
```diff
@@ -349,12 +349,41 @@ function mixinLifecycleEvents(target) {
 }
 
 // TODO: support injection somehow as well?
-export const Observer = observer(({ children }) => children())
+export const Observer = observer(({ children, inject: observerInject, render }) => {
+    const component = children || render
```
FredyC: I am thinking if there shouldn't be some warning if both props are used. Perhaps that's what you had in mind with that propTypes check? Except you need to return message describing a problem. Returning undefined means it's valid.

Sunshine168: ennn , thinging,  first i prefer show warning by propTypes if both props 
and coding like this 
```
const ObserverPropsCheck = (props,key,componentName,location,propFullName)=>{
    const extraKey = key === "children" ? "render" : "children"
    if(typeof propValue[key] === "function" && typeof propValue[extraKey] === "function" ){
            return new Error("Invalid prop,do not use children and render in the same time in`" + componentName )
    }
    
    if (typeof propValue[key] === "function" || typeof propValue[extraKey] === "function") {
        return
    }
    return new Error(
        "Invalid prop `" +
            propFullName +
            "` of type `" +
            typeof propValue[key] +
            "` supplied to" +
            " `" +
            componentName +
            "`, expected `function`."
    )
}

```

'>< but seem complexity     
or do same props check in Observer Component  instead in propType

FredyC: Yea, checking with prop types like this is probably enough. I don't think it adds complexity, it's a friendly behavior toward user.

Sunshine168: great !

README.md
```diff
@@ -105,6 +103,53 @@ React.render(<App person={person} />, document.body)
...
+Example
+
+```javascript
+class App extends React.Component {
+  render() {
+     return (
+         <div>
+            {this.props.person.name}
+            <Observer
+              render={(props) => <div>{props.person.name}</div>}
```
Gvozd: `props` work without inejction?

FredyC: Good catch, this is an oversight. It should refer to `this.props` instead.

Sunshine168: Sorry ,a stupid fault  zzzzzz    should refer to this.props

mweststrate:
Great work @Sunshine168 and thanks for the extensive reviews @FredyC and @Gvozd ! Will release tomorrow

mvestergaard:
This is broken. Results in `Warning: Failed prop type: propValue is not defined` errors due to usage here https://github.com/mobxjs/mobx-react/blob/master/src/observer.js#L368

FredyC:
@mvestergaard That's an unfortunate oversight. Could you send in the PR please? Some tests for this should be probably included as well.

mweststrate:
Just released 4.4.1, that should fix the issue. Unclear yet why this passed the tests, so if somebody could add a test testing the prop types checking that would be appreciated :D

PR #407, opened 2018-01-31
Add husky and run Prettier manually

FredyC:
For some reason `husky` wasn't installed so `precommit` hook for running Prettier was never used. That leads to PR like #403 which also includes reformatting so it's harder to review.

I've run Prettier manually to fix formatting in all files. I've also added `.prettierignore` for `package.json` which is also formatted differently by Yarn on any install.

Lastly there is no need to repeat Prettier rules in for `lint-staged`, the `.prettierrc` file is respected correctly.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,1 @@
+package.json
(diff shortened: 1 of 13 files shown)


mweststrate:
Thanks!

PR #411, opened 2018-02-03
Support correct propagation of changed/replaced stores

Gvozd:
That fix problem, demonstrated in test "warning is printed when changing stores"


Gvozd:
`context.mobxStores` at now in mobx.computed
Provider observe parent context and own props, like in ReactiveMixin, and produce computed mobxStores
Injector at now is observer-decorated, and observe mobxStores

mweststrate:
Hey @Gvozd, thanks for the effort. But please open an issue first to discuss a core change to a package. That might potentially safe a lot of work. 

I have some doubts about this PR, as, if I interpret the change correctly, with the above change it is still not possible to inject new things in a component if there is a `false` returning `sCU in one of it's parents. (One could argue that the warning should probably be an error). I think warning and supporting it will just add confusion; as if it does work, people will want to disable the warning, but if the warning is disabled, it might break in other places in the same app. 

Also from a timeline perspective I am not sure whether this intermediate step adds value, as we can just move to the new React context api in a next major, and simply support changing stores without any further problems.

Gvozd:
@mweststrate I opened #414

mweststrate:
Perfect, will check in a few weeks as this one requires some tinkering and
already tinkering about other stuff😊

Op wo 7 feb. 2018 20:42 schreef Gvozd <notifications@github.com>:

Issue #415, opened 2018-02-07
mobx-react .babelrc causing issues when running jest


IljaDaderko:
I'm encoutering an error when running my jest tests, that I believe is related to `transform-decorators-legacy` used in `.babelrc` for mobx-react, error is as follows

> ReferenceError: Unknown plugin "transform-decorators-legacy" specified in "/Users/me/Documents/Repositories/project/node_modules/mobx-react/.babelrc.env.test" at 0, attempted to resolve relative to "/Users/me/Documents/Repositories/project/node_modules/mobx-react"

Sunshine168:
try ```npm install babel-plugin-transform-decorators-legacy```

emmm  i guess babelrc should be added to the .npmignore ... Try remove .babelrc file in mobx-react to help me prove it  will be nice

IljaDaderko:
@Sunshine168 I think `babel-plugin-transform-decorators-legacy` will be pouting my repo, as I don't use it. Your suggestion to .npmignore / delete `.babelrc` is probably better. I confirm that deleting it from node_modules/mobx-react fixes this 👍

ryanmr:
Hey there. Thanks @IljaDaderko for making this issue. Over the last few days, I've also been receiving an error from our `jest` tests. It's different than the one mentioned above, but I wanted to document it in case anyone else runs into it.

> ● Test suite failed to run
>    Couldn't find preset "es2015" relative to directory 
>".../node_modules/mobx-react"
>      at node_modules/babel-core/lib/transformation/file/options/option-manager.js:293:19
>          at Array.map (<anonymous>)

I can confirm that _deleting_ `.babelrc` file fixes this error. Adding that to `. npmignore` would likely be the best bet.

anushbmx:
Adding .babelrc to .npmignore solves the problem or remove it for now 
`rm ./node_modules/react-native/mobx-react/.babelrc`

mweststrate:
Shipped the solution by @ryanmr in 4.4.2. Thanks for investigating! Could you confirm the fix @IljaDaderko ?

mweststrate:
Does mobx-react@4.4.2-fix-428 fix the issue as well?

mweststrate:
Assuming fixed for now :)

PR #416, opened 2018-02-07
Adds .babelrc to .npmignore

ryanmr:
Fixes issue https://github.com/mobxjs/mobx-react/issues/415

Thank you for the consideration!


mweststrate:
Thanks, nice catch!

mweststrate:
Released as 4.4.2

Issue #419, opened 2018-02-11
Using props on an <Observer> component


carlosagsmendes:
Hi,

I have a very simple functional component where I'm using `inject` and `observer`:

```
const MobxFunctionalComponent = inject("store")(
  observer(({ store, title }) => (
    <div>
      {title} {Date.now()}
      &nbsp;{store.color}
    </div>
  ))
);
```

I'm trying to move away from the decorators and use the <Observer> component instead. But the `title` property is no longer working and the component is being re-rendered even when I don't update the `title` and `color` props:

```
const MobxFunctionalComponentObserver = () => (
  <Observer
    inject={(stores) => ({ store: stores.store })}
    render={props => (
      <div>
        {props.title} {Date.now()}
        &nbsp;{props.store.color}
      </div>
    )}
  />
)
```

I'm a newbie to react and mobx and must be missing something. Can you help? 

Here's the link to an online [example on CodeSandBox](https://codesandbox.io/s/k378jn3817)

Sunshine168:
emmm  i feel sorry 
it is a bug because https://github.com/mobxjs/mobx-react/blob/master/src/observer.js#L360
i will fix it soon

urugator:
Following works, but it's a bit confusing...
```javascript
const MobxFunctionalComponentObserver = props => ( // this is where "title" originally comes from
  <Observer
    inject={(stores) => ({ title: props.title, color: stores.store.color })} // inject provides "props" for "render" so we have to pass the "title" here
    render={props => ( // "props" returned from "inject"
      <div>
        {props.title} {Date.now()}
        &nbsp;{props.color}
      </div>
    )}
  />
);
```

Sunshine168:
emm  confusion in mine new test suit  


```js
// test1
    test("use render with inject should be correct", async () => {
        const person  = mobx.observable({
            first:'hello',
            second:'world',
        })
        const Comp = () => (
            <div>
                <Observer
                    inject={['person']}
                    render={props => <span>{`${props.person.first} ${props.person.second}`}</span>}
                />
            </div>
        )
        const A = () => (
            <Provider person={person}>
                <Comp />
            </Provider>
        )
        await asyncReactDOMRender(<A />, testRoot)
        expect(testRoot.querySelector("span").innerHTML).toBe("hello world")
        person.first = 'hi'
        await sleepHelper(100)
        expect(testRoot.querySelector("span").innerHTML).toBe("hi world")
    })
   // test 2
[...]

PR #420, opened 2018-02-11
Fix Observer with inject and update the correct test suit

Sunshine168:
#419


urugator:
It wraps the component twice inside `observer` that shouldn't be necessary I think...

Sunshine168:
i  have the same thought before 
the look at https://github.com/mobxjs/mobx-react/issues/419#issuecomment-364754426  
i feel confused  ,  want to investigate the reason

PR #431, opened 2018-03-04
Updated readme to include information about potential lifecycle conflicts

JaxCavalera:
As discussed with @mweststrate there's a gotcha that can sneak in when using arrow functions for React lifecycle hooks so this PR adds a small note with link to relevant issue to the docs to help other users that run into this situation.


mweststrate:
Thanks!

PR #434, opened 2018-03-08
Smart inject typing

### LINKED ISSUE #256: [Question] How to get typesafe injection
I've looked at
https://github.com/mobxjs/mobx-react#strongly-typing-inject

but am not sure if there is a better way. I generally pick out a lot of things to inject. Imagine

```
@inject((stores) => {
  return {
    storeA: stores.storeA,
    storeB: stores.storeB,
    storeC: stores.storeC,
    somethingOnStoreD: stores.storeD.something
})
MyComponent extends React.Component<IMyComponentProps>
```

But if I render this component from a parent component, then I can't render it without marking each one of the stores or props as optional in IMyComponentProps or pass in values for each of those props. This is documented but makes it very difficult to actually use the props in the component.

Basically anytime I use something from storeA, storeB, or somethingOnStoreD - I first have to check for the presence of those or else typescript will complain that it is potentially undefined.

I think part of the problem is typescript currently doesn't allow class mutations from decorators. But even when I use inject as a function, the tsd file specifies inject returns the original target which expects the props declared in IMyComponentProps
`export function inject<T, P>(storesToProps : IStoresToProps<T, P>): (<TFunction extends IReactComponent<T | P>>(target: TFunction) => (TFunction & IWrappedComponent<T>)); // decorator`

If I were to change it the tsd to 

`export function inject<T, P>(storesToProps : IStoresToProps<T, P>): (<TFunction extends IReactComponent<T | P>>(target: TFunction) => (IReactComponent<P> & IWrappedComponent<T>)); // decorator`

and inject with

```
interface IPropsFromStore {
  storeA: storeA;
  storeB: storeB;
  ...
}
IPropsFromParent {
}
inject<IPropsFromStore, IPropsFromParent>((stores) => {
  return {
    storeA: stores.storeA,
    storeB: stores.storeB,
    storeC: stores.storeC,
    somethingOnStoreD: stores.storeD.something
})(
MyComponent extends React.Component<IPropsFromStore & IPropsFromParent>)
```

This allows me to use the function to return a component that no longer requires all the props that will be inserted with injection. It also will validate that the inject function is returning the props that I am expecting and also allows MyComponent to use props from injection directly and force checks on props from parent. The problem is this declaration will no longer allow inject to be used as a decorator since its no longer clear that inject returns a signature compatible with the original constructor. I kind of wish this functionality somehow existed because I could be willing to forego decorators to have a better type-checking experience but not sure if its possible.

Sorry this was a bit long, do you know of a better way to solve these problems? 

Thanks!

originerd:
Fixes #256.

```ts
interface Store {
  profile: {
    name: string;
  };
}

interface InjectProps {
  name: string;
}
  
interface OwnProps {
  age: number;
}

type Props = InjectProps & OwnProps;

// way 1
inject<Stores, Props, InjectedProps>((stores) => ({
  name: stores.profile.name,
})(ProfileComponent);

// way 2
inject((stores: Stores, props: Props): InjectProps => ({
  name: stores.profile.name,
})(ProfileComponent);

// way 3 (if Props is same as InjectProps)
inject((stores: Stores): InjectProps => ({
  name: stores.profile.name,
})(ProfileComponent);
```

In this case, it needs only the props which are not provided by `inject` function to render injected component.

But, there is a problem that it doesn't support `decorator`.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,4 +1,10 @@
 /**
+ *  Diff & Omit taken from https://github.com/Microsoft/TypeScript/issues/12215#issuecomment-311923766
+ */
+type Diff<T extends string, U extends string> = ({[P in T]: P } & {[P in U]: never } & { [x: string]: never })[T];
+type Omit<T, K extends keyof T> = Pick<T, Diff<keyof T, K>>
+
+/**
  * Turns a React component or stateless render function into a reactive component.
  */
 import React = require("react")
@@ -40,9 +46,12 @@ export type IWrappedComponent<P> = {
 export function inject(
     ...stores: string[]
 ): <T extends IReactComponent>(target: T) => T & IWrappedComponent<T>
-export function inject<S, P, I, C>(
+export function inject<S, P extends I, I>(
+    fn: IStoresToProps<S, P, I>
+): <T extends IReactComponent>(target: T) => IReactComponent<Omit<P, keyof I>> & IWrappedComponent<T>
+export function inject<S, P extends I, I, C>(
     fn: IStoresToProps<S, P, I, C>
-): <T extends IReactComponent>(target: T) => T & IWrappedComponent<T>
+): <T extends IReactComponent>(target: T) => IReactComponent<Omit<P, keyof I>> & IWrappedComponent<T>
 
 // Ideal implemetnation:
 // export function inject


mweststrate:
@originerd what is the status of this PR? Note that the tests are still failing, and ideally they shouldn't to make sure the dev experience is good even when the generic args are not spelled out by the dev.

```

> mobx-react@4.4.3 test:ts /home/travis/build/mobxjs/mobx-react
> tsc -p test/ts
test/ts/compile-ts.tsx(95,1): error TS1238: Unable to resolve signature of class decorator when called as an expression.
  Type 'StatelessComponent<Pick<{ x: number; }, never>> & IWrappedComponent<typeof T11>' is not assignable to type 'void | typeof T11'.
    Type 'StatelessComponent<Pick<{ x: number; }, never>> & IWrappedComponent<typeof T11>' is not assignable to type 'typeof T11'.
      Type 'StatelessComponent<Pick<{ x: number; }, never>> & IWrappedComponent<typeof T11>' provides no match for the signature 'new (props?: { pizza: number; x?: number | undefined; } | undefined, context?: any): T11'.
test/ts/compile-ts.tsx(114,18): error TS2339: Property 'pizza' does not exist on type 'IntrinsicAttributes & Pick<{ x: number; }, never> & { children?: ReactNode; }'.
test/ts/compile-ts.tsx(117,18): error TS2339: Property 'pizza' does not exist on type 'IntrinsicAttributes & Pick<{ x: number; }, never> & { children?: ReactNode; }'.
test/ts/compile-ts.tsx(160,33): error TS2339: Property 'x' does not exist on type 'IntrinsicAttributes & Pick<{}, never> & { children?: ReactNode; }'.
test/ts/compile-ts.tsx(163,1): error TS1238: Unable to resolve signature of class decorator when called as an expression.
  Type 'StatelessComponent<Pick<{}, never>> & IWrappedComponent<typeof LoginContainer2>' is not assignable to type 'void | typeof LoginContainer2'.
    Type 'StatelessComponent<Pick<{}, never>> & IWrappedComponent<typeof LoginContainer2>' is not assignable to type 'typeof LoginContainer2'.
      Types of property 'contextTypes' are incompatible.
        Type 'ValidationMap<any> | undefined' is not assignable to type 'ValidationMap<any>'.
          Type 'undefined' is not assignable to type 'ValidationMap<any>'.
```

Issue #380, opened 2017-12-24
Using `props` causes reactions and computeds to trigger in undesirable ways


iEchoic:
Hello,

We've been using mobx reactions/computeds/etc. in what I thought was an idiomatic way, but I'm now finding that we're triggering a lot of unexpected reactions (and probably computed executions as well) when referencing `this.props`.  Here's the issue:

```
@observable someObservable = 'test';

// this autorun will change every time *any* props change, causing lots
// of excessive calls
autorun(() => {
  this.props.onChange(this.someObservable);
})

// this autorun will only change when `this.someObservable` or `this.props.onChange`
// changes
autorun(() => {
  const onChange = expr(() => this.props.onChange);
  onChange(this.someObservable);  
})
```

The only solution I see is to create a `computed` or use `expr` for every single prop usage.

I believe that this is behavior is by design in current mobx-react. If I'm understanding correctly, this is happening because `this.props` itself (as opposed to only the values on `props`) is observable, so every time props changes, it triggers reactions/computations. 

This behavior is unexpected to me. One of the reasons I love mobx is that if you write idiomatic mobx, everything magically computes/reacts a minimal number of times. However, the first example, this feels like an idiomatic way to use reactions and `prop`s to me, yet it behaves unexpectedly. Is there a better solution here, or some way we can make `props` with `autorun`/`computed`/etc. behave in a less-unexpected way?

urugator:
It's impossible to make individual props observable, see #347
However I think the main problem here is that `onChange` handler is passed as anonymous inline function, therefore the "pure render" check always fails and props are always considered modified, which in turn triggers autoruns (to avoid staleness, when props are actually updated).

iEchoic:
We never use inline anonymous functions for onChange handlers - it’s actually that other props are changing. I agree with @adamabeshouse in that thread that the ideal behavior would be that individual props are treated as observable (even if there’s some intermediate hack to make them appear to behave that way). Is the proposal at the end of that thread not possible/is “impossible” the right term here?

In #347 the example was somewhat non-idiomatic mobx, but the example above is, so I think this is a symptom that the observable behavior of props isn’t quite right. Is the issue purely an implementation problem, and we agree on the ideal behavior of individual props being observable? I can’t tell from that issue. I’d be interested in finding solutions to it if so.

urugator:
> the ideal behavior of individual props being observable

[...]

Issue #347, opened 2017-10-18
observer: props is, itself, observable when only each prop should be


adamabeshouse:
EDITED because original query was not coherent:

[In this jsfiddle](https://jsfiddle.net/xpd2remq/1/), there's a toy model of a scenario I'm facing: I have multiple subcomponents in `render` which I want to not re-compute unnecessarily, and I was hoping to be able to use the observable nature of props together with the Observer tag to implement this. However, because [its not the properties of props which are tracked, but the reference to `this.props` itself](https://github.com/mobxjs/mobx-react/blob/master/src/observer.js#L143), both subcomponents will recompute every time any props change, regardless of which props are actually used in each subcomponent.

The class in the fiddle is duplicated below:
```
@observer 
class Person extends Component<{ firstName:string, lastName:string }, {}> {
    constructor(props) {
        super(props);
        this.getFirstName = ()=>{
            return (<span>{this.props.firstName}</span>);
        };
        this.getLastName = ()=>{
            return (<span>{this.props.lastName}</span>);
        };
    }
    render() {
        return (
          <div>
              <Observer>{this.getFirstName}</Observer>
              <Observer>{this.getLastName}</Observer>
          </div>
        );
    }
}
```

There are two possible solutions to regain the desired functionality:
(1) Use components (decorated with `@observer`), and pass in desired data through props, in order to regain the shallow comparison functionality in `shouldComponentUpdate`, e.g.
```
@observer
class FirstName extends Component<{name:string},{}>{

...

<FirstName name={this.props.firstName}/>
```

(2) Wrap all prop accesses in `@computed`'s, e.g.
```
 this.getFirstName = ()=>{
            return (<span>{this.firstName}</span>);
};


@computed get firstName() {
    return this.props.firstName;
}
``` 

Both are more unwieldy than if the props object were observable, thus each prop tracked individually, instead of the reference to the props object.

mweststrate:
Could you setup a fiddle / sandbox demonstrating this issue?

Op do 19 okt. 2017 om 00:35 schreef Adam Abeshouse <notifications@github.com

adamabeshouse:
Sorry my original post was actually wrong, I should have made a fiddle first.

https://jsfiddle.net/qqsow33p/5/ shows the situation. In fact in both cases I included above, it would be rendered twice. The only time it would only be rendered once is if the <span> in render() is wrapped by an <Observer>, AND the prop access is wrapped by a @computed.

[...]

PR #436, opened 2018-03-09
Optimize reactive props

Strate:
Particular fix for #380 #347

Achieve granular observale state/props

Technique:
Instead of having only 1 atom for whole props/state, we have N atoms for each property/state element, and extra one which track keyset change

When new props object set, we detect key change and report changed for extra atom. Also we instrument props/state object and in-place change properties to getter+setter, which uses atoms and single value storage. Also, previous props/state object should be freezed with it values and de-instrumented from getters and setters, to get lifecycle methods work.

Having extra atom, which detects keys change helps us to detect new props/state element appears.

~~Unfortunatelly, to make in-place instrument with props/state object, I forced to make a extra wrapper for Observer component, which calls to `createElement` and un-freeze it props. This is definetely not good for performance, but `React.createElement` freeze objects :( I think this extra wrapper get this issue back: https://github.com/mobxjs/mobx/issues/405 (and many tests are failed because of this)~~
UPD: solution found

Comments are requested.

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -110,8 +110,6 @@ function is(x, y) {
     }
 }
 
-const skipRenderKey = Symbol("skipRender")
-
 /**
  * ReactiveMixin
  */
--- a/src/observer.js
+++ b/src/observer.js
@@ -1,6 +1,6 @@
 import React, { Component, PureComponent } from "react"
 import hoistStatics from "hoist-non-react-statics"
-import { createAtom, Reaction, _allowStateChanges } from "mobx"
+import { createAtom, Reaction, _allowStateChanges, runInAction } from "mobx"
 import { findDOMNode as baseFindDOMNode } from "react-dom"
 import EventEmitter from "./utils/EventEmitter"
 import inject from "./inject"
@@ -71,14 +71,14 @@ function patch(target, funcName, runMixinFirst = false) {
     const f = !base
         ? mixinFunc
         : runMixinFirst === true
-          ? function() {
-                mixinFunc.apply(this, arguments)
-                base.apply(this, arguments)
-            }
-          : function() {
-                base.apply(this, arguments)
-                mixinFunc.apply(this, arguments)
-            }
+            ? function() {
+                  mixinFunc.apply(this, arguments)
+                  base.apply(this, arguments)
+              }
+            : function() {
+                  base.apply(this, arguments)
+                  mixinFunc.apply(this, arguments)
+              }
 
     // MWE: ideally we freeze here to protect against accidental overwrites in component instances, see #195
     // ...but that breaks react-hot-loader, see #231...
@@ -122,8 +122,8 @@ function makeComponentReactive(render) {
         const individualValues = {}
 
         /**
-             * In-place convert properties to getter+setter
-             */
+         * In-place convert properties to getter+setter
+         */
         function convertStorageToReactive(storage) {
             if (null == storage) {
                 return storage
@@ -137,7 +137,7 @@ function makeComponentReactive(render) {
             }
             Object.keys(storage).forEach(function(key) {
                 if (!(key in individualAtoms)) {
-                    individualAtoms[key] = new Atom("this." + propName + "." + key)
+                    individualAtoms[key] = createAtom("this." + propName + "." + key)
                 }
                 const currentKeyValue = storage[key]
                 delete storage[key]
@@ -161,8 +161,8 @@ function makeComponentReactive(render) {
         }
 
         /**
-             * In-place convert getter+setter to plain property
-             */
+         * In-place convert getter+setter to plain property
+         */
         function convertStorageToStatic(storage) {
             if (null == storage) {
                 return
--- a/src/observer.js
+++ b/src/observer.js
@@ -1,6 +1,6 @@
 import React, { Component, PureComponent } from "react"
 import hoistStatics from "hoist-non-react-statics"
-import { createAtom, Reaction, _allowStateChanges, runInAction } from "mobx"
+import { createAtom, Reaction, _allowStateChanges, runInAction, observable } from "mobx"
 import { findDOMNode as baseFindDOMNode } from "react-dom"
 import EventEmitter from "./utils/EventEmitter"
 import inject from "./inject"
@@ -118,14 +118,13 @@ function makeComponentReactive(render) {
         let currentValue = this[propName]
         let currentValueKeys = null == currentValue ? [] : Object.keys(currentValue)
         const changeKeysAtom = createAtom("this. " + propName + " [change keys]")
-        const individualAtoms = {}
-        const individualValues = {}
+        const propBoxes = {}
 
         /**
          * In-place convert properties to getter+setter
          */
         function convertStorageToReactive(storage) {
-            if (null == storage) {
+            if (!storage) {
                 return storage
             }
             // props object is frozen, so we need to clone it
@@ -136,22 +135,22 @@ function makeComponentReactive(render) {
                 storage = Object.assign({}, storage)
             }
             Object.keys(storage).forEach(function(key) {
-                if (!(key in individualAtoms)) {
-                    individualAtoms[key] = createAtom("this." + propName + "." + key)
+                if (!(key in propBoxes)) {
+                    propBoxes[key] = observable.box(void 0, {
+                        deep: false,
+                        name: "this." + propName + "." + key
+                    })
                 }
                 const currentKeyValue = storage[key]
-                delete storage[key]
                 Object.defineProperty(storage, key, {
                     configurable: true,
                     enumerable: true,
                     get: function get() {
-                        individualAtoms[key].reportObserved()
-                        return individualValues[key]
+                        return propBoxes[key].get()
                     },
                     set: function set(v) {
-                        if (individualValues[key] !== v) {
-                            individualValues[key] = v
-                            individualAtoms[key].reportChanged()
+                        if (propBoxes[key].get() !== v) {
+                            propBoxes[key].set(v)
                         }
                     }
                 })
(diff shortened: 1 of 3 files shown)


Strate:
Hmm, seems we can optimize things, if report to keysAtom only on new key appears, but not on old rey removed.

Strate:
Found solution without extra component wrapper.
Done with clone-if-frozen technique, and this plays well with both state and props.
And all tests are good.

We can make further optimization, by tracking by extra atom only new keys, not removing, but this requires more work, and I don't know worth it is or not.

I think that there could be done some optimizations in code which make storage observable/static.

And, this code could be easily adopted to Proxy in future.

I will test this patch on my project and give some feedback in next week.

src/observer.js
```diff
@@ -138,26 +138,104 @@ const reactiveMixin = {
...
+                    const currentKeyValue = storage[key]
+                    delete storage[key]
+                    Object.defineProperty(storage, key, {
+                        configurable: true,
+                        enumerable: true,
+                        get: function get() {
+                            individualAtoms[key].reportObserved()
+                            return individualValues[key]
+                        },
+                        set: function set(v) {
```
Strate: I think we can make property read-only (by defining only `get`), this can help to make props read-only too (BC with current react behaviour)

mweststrate:
Hey @Strate , thanks! Will try to review in a few days

Strate:
Caveat: if we use reaction like this:
```
reaction (() => this.props, props => console.log(props))
```
it will not react as before, it will only react on keys changed (BC break here)

mweststrate:
@Strate that works for normal observable objects neither, so I think that should be ok

src/observer.js
```diff
@@ -138,26 +138,104 @@ const reactiveMixin = {
...
+            let currentValueKeys = null == currentValue ? [] : Object.keys(currentValue)
+            const changeKeysAtom = new Atom("this." + propName + " [change keys]")
+            const individualAtoms = {}
+            const individualValues = {}
+
+            /**
+             * In-place convert properties to getter+setter
+             */
+            function convertStorageToReactive(storage) {
+                if (null == storage) {
```
mweststrate: just `if (!storage)` ?

src/observer.js
```diff
@@ -138,26 +138,104 @@ const reactiveMixin = {
...
+                    return storage
+                }
+                // props object is frozen, so we need to clone it
+                // we can do it safely inside componentWillMount (react won't warn us about it)
+                // btw, state object can not be cloned here, (react warns about direct state modification), however
+                // state object is not frozen, so we do not need to clone it to instrument
+                if (Object.isFrozen && Object.isFrozen(storage)) {
```
mweststrate: Hmm, if we implemented this using Proxies, it would all be lot simpler I guess? I think that could be fine as well, only introduce this feature in the next mobx-react that accompanies MobX 5

src/observer.js
```diff
@@ -138,26 +138,104 @@ const reactiveMixin = {
...
+                    storage = Object.assign({}, storage)
+                }
+                Object.keys(storage).forEach(function(key) {
+                    if (!(key in individualAtoms)) {
+                        individualAtoms[key] = new Atom("this." + propName + "." + key)
+                    }
+                    const currentKeyValue = storage[key]
+                    delete storage[key]
```
mweststrate: skip the delete, will probably make it slower, without adding value as it is immediately redefined

src/observer.js
```diff
@@ -138,26 +138,104 @@ const reactiveMixin = {
...
+                // props object is frozen, so we need to clone it
+                // we can do it safely inside componentWillMount (react won't warn us about it)
+                // btw, state object can not be cloned here, (react warns about direct state modification), however
+                // state object is not frozen, so we do not need to clone it to instrument
+                if (Object.isFrozen && Object.isFrozen(storage)) {
+                    storage = Object.assign({}, storage)
+                }
+                Object.keys(storage).forEach(function(key) {
+                    if (!(key in individualAtoms)) {
+                        individualAtoms[key] = new Atom("this." + propName + "." + key)
```
mweststrate: might be simpler to just create a collection of `observable.box` -es? That combines an atom with a value holder

src/observer.js
```diff
@@ -138,26 +138,104 @@ const reactiveMixin = {
...
+                    storage[key] = currentKeyValue
+                })
+                return storage
+            }
+
+            /**
+             * In-place convert getter+setter to plain property
+             */
+            function convertStorageToStatic(storage) {
+                if (null == storage) {
```
mweststrate: !storage

src/observer.js
Strate: Unfortunatelly, we can't wrap state to proxy here, because react will check this and warn (proxied object !== original object). But for frozen props, of course, it could be applied.

Strate:
@mweststrate 

> that works for normal observable objects neither, so I think that should be ok

Yes, but right now it works differently for props & state, and this would be BC break. It hink we can just note this in UPGRADE guide

src/observer.js
Strate: Good catch, of course boxes would be better

mweststrate:
Hey @Strate, were you going to process the above comments?

If you could rebase it on the `react16_3` branch that would be great, but if it is too invasive, don't bother

Strate:
Hey @mweststrate 
Just started to work on it.

Strate:
@mweststrate merged with master, also processed all above comments.

Strate:
Maybe it would be better to move new mobx4 object api? :thinking:

Strate:
@xaviergonz for what reason?

xaviergonz:
oh, if you don't want to break backwards compatibility

Strate:
@xaviergonz yes, it works. There is a test for that case.

xaviergonz:
@Strate awesome, I can't wait to see it on mobx itself 👍

Strate:
Unfortunatelly, this PR is break now, because of #496

Strate:
So, we have a problem here. To get this PR work, I need to instrument props passed to component, but I can't do this because react warns that props are modified. I'm not sure that it is possible at all.

xaviergonz:
can you pass to the constructor a modified props?

Strate:
@xaviergonz I tried that solution for a while, but some tests still failing. And this solution need an extra wrapper for a component (which is performance gain)

hong1997:
So finally, is PR ‘Achieve granular observale state/props’ implemented in newest release now?

bsmith-cycorp:
Is this still alive? Is the React warning a total roadblock?

Strate:
@bsmith-cycorp @hong1997 this PR is still pending, because of that warning. That warning could not be just ignored, because this can break some stuff (and it definetely break, because tests are failing in this PR)

xaviergonz:
I do something like this to get reactive props

```ts
import { extendObservable, isObservableProp, observable, reaction } from "mobx";
import { disposeOnUnmount } from "mobx-react";
import * as React from "react";

export abstract class ObsComponent<P = {}, S = {}> extends React.Component<P, S> {
  @observable.shallow
  obsProps = this.props;

  constructor(props: P, context?: any) {
    super(props, context);
    const obsProps: any = this.obsProps;

    disposeOnUnmount(
      this,
      reaction(
        () => this.props,
        (unobsProps: any) => {
          const extend: any = {};
          let extended = false;

          for (const propName of Object.keys(unobsProps)) {
            if (isObservableProp(obsProps, propName)) {
              obsProps[propName] = unobsProps[propName];
            } else {
              extended = true;
              extend[propName] = unobsProps[propName];
            }
          }

          if (extended) {
            extendObservable(obsProps, extend, undefined, { deep: false });
          }
        },
        {
          fireImmediately: true
        }
      )
    );
  }
}
```

Then I extends from ObsComponent instead of Component and use reactions over this.obsProps rather than this.props

Although now that I think about it it would probably be better if it was transformed into a function like
```
class C extends Component {
  this.obsProps = observableProps(this);
}
```

I can make such function and add it to a PR if somebody is interested

xaviergonz:
The good thing of that approach is that it is not gonna break anything :)

Strate:
@xaviergonz yes, it extends observable, and if you will read prop __after__ extendObservable, everything gonna be okay.

Imagine this case:
```
class SomeComp extends ObsComponent {
  @comptued
  get someComputed() {
    return this.obsProps.prop ? 1 : 2;
  }
  render() {
    if (1 == this.someComputed) {
      return "One";
    }
    return "Two";
  }
}
```

Component initially renders as `<SomeComp />`, and then re-renders as `<SomeComp prop={true} />`.
In this case, your computed initially tries to get access to non-existence property, and will not re-compute after property apperas, even if `extendObservable` used.

If you get rid of computed in that case everything gonna to be oka, because react re-renders component if new props come.

xaviergonz:
Hmm, could be used by proxies (but I guess that's the same mobx5 is doing internally and that's why it works)

@mweststrate is there a way to mark an observed property as "dirty" or "invalidate" it?
If so, for each new (non previosly observable) property it could run something like "invalidateProperty(obsProps, propName)" to let the computed function it should look again

Strate:
@xaviergonz in this PR I solved that kind of problem by adding extra atom, which tracks keys change. You can check this PR for implementation details.

Strate:
And yes, there is many ways to achieve granular props, easies is use `expr` helper:
```
class SomeComp extends ObsComponent {
  @comptued
  get someComputed() {
    return expr(() => this.obsProps.prop) ? 1 : 2;
  }
  render() {
    if (1 == this.someComputed) {
      return "One";
    }
    return "Two";
  }
}
```
but main goal of this PR is to get optimized behavior out-of-the-box, without need to rewrite whole app.

mweststrate:
Closing, as this gets less, not more relevant, with the direction hooks are taking

Strate:
@mweststrate could you please share some more info why this is no more relevant?

bsmith-cycorp:
For those landing on this ticket, here's another workaround:

```javascript
function createReactiveProps<T>(componentInstance: { props: T }): T {
	let props = observable({});
	
	Object.assign(props, componentInstance.props)
	autorun(() => Object.assign(props, componentInstance.props))
	return props;
}
```
Usage:
```javascript
class MyComponent extends React.Component<Props> {

  @observable reactiveProps = createReactiveProps(this);
  ...
```

PR #445, opened 2018-03-23
Added a small sample for onError to README.md

### LINKED ISSUE #405: Document using onError
https://github.com/mobxjs/mobx-react#global-error-handler-with-onerror

An example would be nice :)

JoshuaKGoldberg:
Fixes #405


mweststrate:
Thanks!

PR #470, opened 2018-05-04
Fix Component displayed as 'undefined' in mobx-devtools

MauricioAndrades:
I read through entire codebase and this seems to be spot to fix this little bug. Let me know if I'm way off so I can find the right place to patch this.


mweststrate:
Looks proper, thanks!

PR #472, opened 2018-05-07
fixes HOC displayName for observer and inject

### LINKED ISSUE #466: inject HOC display name
Before running around creating PR's everywhere I thought I'd ask why the inject HOC has a non-standard display name wrapper? https://github.com/mobxjs/mobx-react/blob/ab90ae65991c95da73406accd1048e225e63f02e/src/inject.js#L37-L43

Currently this means when we debug we get a wrapper like `inject-Component` or `inject-Component-with-storeA`. This is inconsistent with the rest of the HOC components like `observer(Component)`.

Why isn't inject using displayName like `inject(Component)` or `inject-with-storeA(Component)`?

SimeonC:
Uses the convention for HOC display names as outlined in the React documentation.
https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

Closes #466 <- See for more details.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -34,13 +34,14 @@ const proxiedInjectorProps = {
  * Store Injection
  */
 function createStoreInjector(grabStoresFn, component, injectNames) {
-    let displayName =
-        "inject-" +
-        (component.displayName ||
-            component.name ||
-            (component.constructor && component.constructor.name) ||
-            "Unknown")
-    if (injectNames) displayName += "-with-" + injectNames
+    let displayName
+    const componentName =
+        component.displayName ||
+        component.name ||
+        (component.constructor && component.constructor.name) ||
+        "Component"
+    if (injectNames) displayName = "inject-with-" + injectNames + "(" + componentName + ")"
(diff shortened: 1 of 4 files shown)


mweststrate:
I am not sure about wrapping `observer()` around non-function components; as for normal components it rather acts as mixing then as HoC. (And I think it will be make the entire component tree probably quite unnecessary very verbose?)

The inject names are definitely better this way. 

I marked it as breaking change, as I think it could impact test suites relying on snapshots quite a bit. Not really sure whether that counts as 'breaking' though as there should be no runtime impact. But probably this is best shipped as separate minor at least

SimeonC:
Yea, it probably is a breaking change for that reason.
My thinking adding the HOC convention to `observer` was that it would be useful to see which components would be reacting to mobx changes which currently isn't very visible. Didn't think about whether it would be too verbose with it in but it has caught me out every now and then - it also would help to highlight the difference between `inject('store')` and `inject(() => {})` which only one of these applies observer and the other doesn't.

mweststrate:
Merged into the v6 branch

PR #476, opened 2018-05-17
Don't warn about shouldComponentUpdate from @observer'd superclass

kmdavis:
This pull fixes the issue where the following is logged erroneously:

> Use `shouldComponentUpdate` in an `observer` based component breaks the behavior of `observer` and might lead to unexpected results. Manually implementing `sCU` should not be needed when using mobx-react.

This warning is misleading and incorrect if the `shouldComponentUpdate` that is being complained about was added by `@observer` itself. e.g.

```js
@observer
class GenericComponent extends Component {}

@observer // should not complain, but it does
class SpecificComponent extends GenericComponent {}
```


mweststrate:
Duplicate of #492, thanks for the effort though :-)

PR #480, opened 2018-05-22
added FAQ entry about reactions inside constructors

xaviergonz:


xaviergonz:
I think this merge can be reverted now that it's no longer the case

mweststrate:
Nice catch, done :)

Issue #494, opened 2018-06-04
'mobx' does not contain an export named '$mobx'


pavsidhu:
After installing mobx and mobx-react, I receive the following error:

```
./node_modules/mobx-react/index.module.js
599:24-29 'mobx' does not contain an export named '$mobx'.
```

I've used this library in the past and have had no issues. I'm unsure if this is an issue on my end or not. Let me know if there's anything I can do to help.

garkin:
Getting the same after updating to `mobx-react@5.2.0`.
```json
"dependencies": {
    "mobx": "^4.3.0",
    "mobx-react": "^5.2.0",
    "webpack": "^4.10.2",
    "webpack-cli": "^3.0.1",
    "webpack-dev-server": "^3.1.4",
}
```

garkin:
Loks related https://github.com/mobxjs/mobx-react/commit/18fc79730c581b4b4348cc6ed4b2f271aeddea91
Dudes, you made release minor version increment dependant on `"mobx": "5.0.0-beta.5"`. Not cool.

mweststrate:
Upgrading mobx to 5.3.1 or downgrading mobx-react to 5.1.0 should solve the issue

mweststrate:
Yeah, this was supposed to use dynamic feature detection, to figure if we
are dealing with MobX 4 or 5, but since the static analysis of webpack
can't handle that this breaks if you are bundling with webpack. Just
released mobx 4.3.1 to make it forward compatible. Doesn't solve the issue
100% (still incompatible with 4.3.0), but I am not sure if there is a real
use case where mobx-react is upgraded, but mobx not? Alternatively I can
retract mobx-react@5.2.0 and re-release it as 6.0.0, but that might
existing builds as well.

Op ma 4 jun. 2018 om 18:39 schreef garkin <notifications@github.com>:

Issue #493, opened 2018-06-04
mobx-react 5.2.0 fails to build with Webpack 3.x (create-react-app)


lvpro:
Failed to compile.

./node_modules/mobx-react/index.module.js
599:24-29 'mobx' does not contain an export named '$mobx'.

mweststrate:
Upgrading mobx to 5.3.1 or downgrading mobx-react to 5.1.0 should solve the issue.

Sorry for the inconvenience! In some build setups combining mobx-react 5.2.0 with MobX <= 4.3.0 will yield this problem.

mweststrate:
Sorry, yes 4.3.1 indeed

Op ma 4 jun. 2018 21:24 schreef Alan Rubin <notifications@github.com>:

PR #495, opened 2018-06-04
Fix prepared for Mobx 5 support

Krivega:
see #493 #494

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -1,11 +1,11 @@
 import React, { Component, PureComponent } from "react"
 import hoistStatics from "hoist-non-react-statics"
-import { createAtom, Reaction, _allowStateChanges, $mobx } from "mobx"
+import * as mobx from "mobx"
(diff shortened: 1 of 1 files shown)


src/observer.js
```diff
@@ -1,11 +1,11 @@
 import React, { Component, PureComponent } from "react"
 import hoistStatics from "hoist-non-react-statics"
-import { createAtom, Reaction, _allowStateChanges, $mobx } from "mobx"
+import * as mobx from "mobx"
```
mweststrate: The problem is that imports like these kill tree-shaking (although that mobx is not the most tree shakeable lib). I'll keep this PR open for a bit, it might be good stop-gap if more people run into the same problem and try to combine mobx-react 5.2+ with mobx 5.3.1-

mweststrate:
Closing as the issue has in the mean time adressed by MobX 4.3.1. Thanks for the PR nonetheless!

PR #496, opened 2018-06-07
Make props & state observable in constructor & componentWillMount

### LINKED ISSUE #478: reactions do not work over props when used in constructor, but they work when used in componentDidMount
React 16.3, latest mobx-react

Today I found a weird issue when I was creating a reaction over props in the component constructor (using TS btw):
```
@observer
class MyC extends React.Component {
  constructor(props) {
    super(props);
    this.r = reaction(() => this.props.open, () => {
      // for some reason this would only run once, but never again when the this.props.open changed
    }, { fireImmediately: true });
  }
}
```

however when I moved the reaction to componentDidMount it all started working...

makes me wonder, maybe react is internally making a copy of the props and swapping them before componentDidMount executes?

or maybe props made observable after the constructor runs?
if so then maybe the decorator should override the super class constructor function and inject the make observable thing on it?

Strate:
Moved instrumentation of state & props to `observer` decorator.
Closes #478

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -20,6 +20,9 @@ let warnedAboutObserverInjectDeprecation = false
 export const componentByNodeRegistry = typeof WeakMap !== "undefined" ? new WeakMap() : undefined
 export const renderReporter = new EventEmitter()
 
+const skipRenderKey = Symbol("skipRender")
+const isForcingUpdateKey = Symbol("isForcingUpdate")
+
 function findDOMNode(component) {
     if (baseFindDOMNode) {
         try {
@@ -116,29 +119,6 @@ function is(x, y) {
 function makeComponentReactive(render) {
     if (isUsingStaticRendering === true) return render.call(this)
 
-    function makePropertyObservableReference(propName) {
-        let valueHolder = this[propName]
-        const atom = createAtom("reactive " + propName)
-        Object.defineProperty(this, propName, {
-            configurable: true,
-            enumerable: true,
-            get: function() {
-                atom.reportObserved()
-                return valueHolder
-            },
-            set: function set(v) {
-                if (!isForcingUpdate && !shallowEqual(valueHolder, v)) {
-                    valueHolder = v
-                    skipRender = true
-                    atom.reportChanged()
-                    skipRender = false
-                } else {
-                    valueHolder = v
-                }
-            }
-        })
-    }
-
     function reactiveRender() {
         isRenderingPending = false
         let exception = undefined
@@ -177,17 +157,12 @@ function makeComponentReactive(render) {
      * If props are shallowly modified, react will render anyway,
      * so atom.reportChanged() should not result in yet another re-render
      */
-    let skipRender = false
+    this[skipRenderKey] = false
     /**
      * forceUpdate will re-assign this.props. We don't want that to cause a loop,
      * so detect these changes
      */
-    let isForcingUpdate = false
-
-    // make this.props an observable reference, see #124
-    makePropertyObservableReference.call(this, "props")
-    // make state an observable reference
-    makePropertyObservableReference.call(this, "state")
+    this[isForcingUpdateKey] = false
 
     // wire up reactive render
     const baseRender = render.bind(this)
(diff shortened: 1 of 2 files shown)


src/observer.js
```diff
@@ -177,17 +157,12 @@ function makeComponentReactive(render) {
      * If props are shallowly modified, react will render anyway,
      * so atom.reportChanged() should not result in yet another re-render
      */
-    let skipRender = false
+    this[skipRenderKey] = false
```
mweststrate: could we make this a non-enumerable prop?

src/observer.js
```diff
@@ -177,17 +157,12 @@ function makeComponentReactive(render) {
...
      * forceUpdate will re-assign this.props. We don't want that to cause a loop,
      * so detect these changes
      */
-    let isForcingUpdate = false
-
-    // make this.props an observable reference, see #124
-    makePropertyObservableReference.call(this, "props")
-    // make state an observable reference
-    makePropertyObservableReference.call(this, "state")
+    this[isForcingUpdateKey] = false
```
mweststrate: idem

src/observer.js
```diff
@@ -273,6 +248,32 @@ const reactiveMixin = {
     }
 }
 
+function makeObservableProp(target, propName) {
+    const valueHolderKey = Symbol(propName + " value holder")
+    const atomHolderKey = Symbol(propName + " atom holder")
+    function getAtom() {
+        return this[atomHolderKey] || (this[atomHolderKey] = createAtom("reactive " + propName))
```
mweststrate: could we make this a non-enumerable prop?

src/observer.js
Strate: I think yes, of course

src/observer.js
Strate: Please clarify, dont understood :(

src/observer.js
```diff
@@ -273,6 +248,32 @@ const reactiveMixin = {
     }
 }
 
+function makeObservableProp(target, propName) {
+    const valueHolderKey = Symbol(propName + " value holder")
+    const atomHolderKey = Symbol(propName + " atom holder")
```
mweststrate: I am wondering whether using Symbols is a breaking change. So far we didn't depend on them, but I think React itself already requires Symbol or a polyfill of it? (Since mobx-react 5 can still be used with mobx 4 so should work on any ES 5 env)

src/observer.js
Strate: Yes. Will address this a bit later today.

mweststrate:
Looks like a great improvement! Left some questions

src/observer.js
Strate: I will check this out, maybe there is a way to do this without symbols.

src/observer.js
mweststrate: Sorry, making this a non-enumerable prop as well :)

src/observer.js
xaviergonz: The support looks pretty good to me: https://caniuse.com/#search=symbol

mweststrate: without symbols it could be done by putting an object under some ugly key. Symbols is preferred though :) You could always create a super naive polyfill like: `function Symbol(name) { return `$mobxReactProp${symbolname}${Math.random()} }`. But since react itself already requires quite some polyfills ([link](https://reactjs.org/docs/javascript-environment-requirements.html]) I think we can bet on Symbol to be present. And otherwise fix it if somebody reports is

Strate:
Made internal props hidden, but keep Symbol as is.

mweststrate:
Merging! Will release soon :-)

Strate:
Wow, cool! Can't wait for release :)

mweststrate:
Released as 5.2.1!

TomaszGrzelak:
@mweststrate - Sadly I've just run into a problem that is 90% likely to be connected with the changes discussed here (got 5.2.1 installed just a few hours ago).
What I found, is that `Symbol` is not natively supported on Android (using React-Native).
I managed to get things working by adding `import 'es6-symbol/implement';` at the very beginning of my index.js file.

Strate:
@TomaszGrzelak oh, sorry for that. I gonna to make PR to naive polyfill Symbol

Strate:
@TomaszGrzelak #499

PR #499, opened 2018-06-08
Polyfill Symbol if not exists

Strate:
React native Android don't have a Symbols. So, polyfill them if not exists


mweststrate:
Great!

mweststrate:
Released as 5.2.2

Strate:
Hmm, is that usage of `global` safe here? Is it transformed to window somehow later? I didnt check this is browser:(

mweststrate:
Afaik yes. And I think it is more likely that `global` exists than that `window` exists on RN.

PR #501, opened 2018-06-09
Get rid of Symbol at all

Strate:
Motivation:

mobx itself has hidden properties and don't use Symbols at all. I don't think that we need symbol here.

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -20,19 +20,16 @@ let warnedAboutObserverInjectDeprecation = false
 export const componentByNodeRegistry = typeof WeakMap !== "undefined" ? new WeakMap() : undefined
 export const renderReporter = new EventEmitter()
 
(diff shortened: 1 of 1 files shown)


mweststrate:
@Strate MobX 5 uses Symbols for the hidden properties, and I think it is the neater approach going forward (no name collisions possible etc). 

I just wasn't sharp enough yesterday :). The proper to check the existence of symbol is:

```
if (typeof Symbol !== "undefined) { 

} else {

}
```

That works everywhere' the `typeof` operator is very funny in JavaScript, as it will not throw if its operand doesn't exist at all, unlike any other JS construction.

Strate:
@mweststrate thank you for that point, just fixed this

mweststrate:
Released as 5.2.3

PR #513, opened 2018-07-02
Make inject like react-redux connect

leejaedus:
I think this way looks better. like react-redux


mweststrate:
@leejaedus please note that the tests are failing. PRs are not processed unless they pass the tests.

mweststrate:
Closing since tests are failing and PR seems inactive

PR #516, opened 2018-07-09
Added docs to explain that state and props are also observables in an…

guiguan:
… observer component


mnpenner:
> `this.props` and `this.state` are also observables themselves

Do you mean they get converted into MobX observables or that they behave like observables in that React [regardless of MobX] will re-render a component when they change?

guiguan:
@mnpenner Although I am not entirely sure how it works underlie, `@computed` also reacts to their changes. A quick glance at the source code, I believe they are made `observables` via `atom`: https://github.com/mobxjs/mobx-react/blob/master/src/observer.js#L277

Issue #479, opened 2018-05-21
Package size engines & dependency on the react-dom


GantMan:
When checking package size on [bundlephobia](https://bundlephobia.com/result?p=mobx-react@5.1.2) I get the following error.

![image](https://user-images.githubusercontent.com/997157/40314397-fff5bfda-5cdd-11e8-9dae-6e244ad45172.png)

I also get an error when I try using the `package-size` module from npm
![image](https://user-images.githubusercontent.com/997157/40314428-129be3da-5cde-11e8-8655-1e333c302a10.png)

I understand this is probably dude to React Native, so you could consider setting an [optional dependency](https://docs.npmjs.com/files/package.json#optionaldependencies)?   I'm not sure if that will work.

mweststrate:
react-dom is not a (peer) dependency because it is only a dependency when using it in a DOM project, but not when using RN, etc.

If bundlephobia tries to resolve react-dom, it sounds like it is picking the wrong entry in the module in the first place, as the RN version doesn't `require('react-dom')`. Probably it doesn't respect this field in the package header? `  "react-native": "native.js",`

mweststrate:
Closing for now, multiple packages would be neat, but the problem doesn't seem urgent enough to invest in it now.

PR #520, opened 2018-07-23
Added react-dom as peer dep

ematipico:
I was checking the size of the package using this website

https://bundlephobia.com/result?p=mobx-react

Although it was complaining about a `react-dom`, which is actually missing as peerDep. Feel free the close the PR if it's not needed

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,8 @@
   "homepage": "https://mobxjs.github.io/mobx",
   "peerDependencies": {
     "mobx": "^4.0.0 || ^5.0.0",
-    "react": "^0.13.0 || ^0.14.0 || ^15.0.0 || ^16.0.0"
+    "react": "^0.13.0 || ^0.14.0 || ^15.0.0 || ^16.0.0",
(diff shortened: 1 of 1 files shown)


mweststrate:
See https://github.com/mobxjs/mobx-react/issues/479 @ematipico; react-dom is not necessarily a peer dependency, only optionally, which is why it isn't specified as one atm

ematipico:
Shouldn't `react-dom` and `react-native` be an optional dependency at least? I mean, they **used** inside the code base. Honest question :)

mweststrate:
Well, actually they should be optional _peer_ dependencies. But afaik that
concept does not exist in npm. Splitting into separate packages (or at
least publishing as separate packages), as suggested by @urugator will
probably solve the issue, as then they can become just normal dependencies.
But since so far the only problem this ever caused seems to be
bundlephobia, not sure whether that would be worth the effort :)

Op ma 23 jul. 2018 om 23:16 schreef Emanuele <notifications@github.com>:

ematipico:
Not at all! Thanks for the clarification!

Issue #500, opened 2018-06-09
Hot reload doesn't work when @observer is used


eyousefifar:
package version:
 "mobx": "^4.3.1",
 "mobx-react": "^5.2.2",
 "react": "16.3.1",
 "react-native": "0.55.4"
platform: Android
Hi, when I add @ observer to my react class, hot reloading doesn't work. it reloads the code but UI doesn't rerender. it works fine without an observer
I can provide more information if needed

urugator:
https://mobx.js.org/best/stateless-HMR.html
https://mobx.js.org/best/decorators.html#disclamer-limitations-of-decorator-syntax
https://github.com/mobxjs/mobx/issues?utf8=%E2%9C%93&q=is%3Aissue+HMR
https://github.com/mobxjs/mobx-react/issues?utf8=%E2%9C%93&q=is%3Aissue+HMR

eyousefifar:
@urugator tnx, but I used class components and removed decorators, and used observer like this : 
export default observer(App) where App is my components class, but as soon as I put observer to work, HMR stops working.

luco:
+1 on this.

theKashey:
@briedel1 - you just spotted one of RHL's limitations. I could understand why it will not work for SFC wrapper by observable (it will not, due to `componentClass` is trapped in function scope), but I don't have explanation why it does not work for classes.

Let me play with it for a while.

theKashey:
So - there are always just 2 issues:
 - something don't get updated. HMR\RHL could not update some things. Re-rendering will not "fix" it.
 - something don't get re-rendered. RHL tries to force-update everything, but that doesn't always work. So - may be HMR/RHL did their job here, just MobX not seeing reason to update component, or visa versa. If it will be re-rendered by the App's property change - that's the case.

Anyway - don't touch anything :), I need a broken example to fix.

briedel1:
Thanks for taking a look.

theKashey:
So everything is badly broken. 
You may enable RHL debug mode using `setConfig`, and then see the problem
>bundle.js:918 React-hot-loader: reconciliation failed due to error TypeError: Cannot read property 'obsObj' of undefined

`this.state` - is really undefined, due to `Symbol('state value holder')` created in `makeObservableProp` is unique.

I've tried to "concrete" them, so allocate one symbol per properly name, and it fixes the problem.

@mweststrate - is there any sense to use unique symbols to shadow "well-known" props like state and props?

[...]

PR #522, opened 2018-07-28
fix: React-Hot-Loader compatibility

theKashey:
Source issue: #500.

Just changing `createSymbol` to behave like pure function and return the same result for the same arguments.
As a result - after hot class update `.state` and `.props` refers the same underlaying objects.
Before - RHL did not work due to runtime error, as long `.state` was `undefined` for a while.


mweststrate:
Thanks!

theKashey:
That PR was a bit incomplete :)

PR #523, opened 2018-07-28
Fix: observer should not modify original componentClass

zhangciwu:
When use as function, observer modifies original componentClass
It could cause `Maximum call stack size exceeded` exception after componentClass's render method got too many wrapping

It happens to me, and I create a repo to reproduce it: [https://github.com/zhangciwu/mobx-react-stack-exceeded](https://github.com/zhangciwu/mobx-react-stack-exceeded)

This PR is to fix it by extending (clone) componentClass before modifying, and keeping original class out of altering.
Also appended some test cases.

### DIFF
--- a/src/observer.js
+++ b/src/observer.js
@@ -368,6 +368,7 @@ export function observer(arg1, arg2) {
     }
 
     const targetClass = class targetClassName extends componentClass {}
+    targetClass.displayName = componentClass.displayName || componentClass.name
     const target = targetClass.prototype || targetClass
     mixinLifecycleEvents(target)
     targetClass.isMobXReactObserver = true
--- a/src/observer.js
+++ b/src/observer.js
@@ -367,16 +367,17 @@ export function observer(arg1, arg2) {
         throw new Error("Please pass a valid component to 'observer'")
     }
 
-    const target = componentClass.prototype || componentClass
+    const targetClass = class targetClassName extends componentClass {}
+    const target = targetClass.prototype || targetClass
     mixinLifecycleEvents(target)
-    componentClass.isMobXReactObserver = true
+    targetClass.isMobXReactObserver = true
     makeObservableProp(target, "props")
     makeObservableProp(target, "state")
     const baseRender = target.render
     target.render = function() {
         return makeComponentReactive.call(this, baseRender)
     }
-    return componentClass
+    return targetClass
 }
 
 function mixinLifecycleEvents(target) {
--- a/test/observer.test.js
+++ b/test/observer.test.js
@@ -838,6 +838,73 @@ describe("use Observer inject and render sugar should work  ", () => {
     })
 })
 
+describe("should not modify original componentClass", () => {
+    test("keep original at client", () => {
+        useStaticRendering(false)
+
+        class Origin extends React.Component {}
+
+        Origin.prototype.render = "DO_NOT_MODIFY"
+        observer(Origin)
+        expect(Origin.prototype.render).toBe("DO_NOT_MODIFY")
+    })
+
+    test("keep original at server", () => {
+        useStaticRendering(true)
+
+        class Origin extends React.Component {}
+
+        Origin.prototype.render = "DO_NOT_MODIFY"
+        observer(Origin)
+        expect(Origin.prototype.render).toBe("DO_NOT_MODIFY")
+    })
+
+    test("no exception: Maximum call stack size exceeded", () => {
+        // this test maybe a little slow (10s)
+
+        useStaticRendering(true)
+
+        class NameDisplayer extends Component {
+            render() {
+                return <h1>{this.props.name}</h1>
+            }
+        }
+
+        function getWraped() {
+            return inject(stores => ({
+                name: stores.userStore.name
+            }))(observer(NameDisplayer))
+        }
+
+        const user = mobx.observable({
+            name: "Noa"
+        })
+
+        class Name2 extends Component {
+            render() {
+                let UserNameDisplayer = getWraped()
+                return (
+                    <div>
+                        <UserNameDisplayer />
+                    </div>
+                )
+            }
+        }
+
+        function getHtml() {
+            return ReactDOMServer.renderToString(
+                <Provider userStore={user}>
+                    <Name2 />
+                </Provider>
+            )
+        }
+
+        for (let i = 0; i < 20000; i++) {
+            getHtml()
+        }
+    })
+})
+
 test("don't use PureComponent", () => {
     const msg = []
     const baseWarn = console.warn


mweststrate:
@zhangciwu thanks for this PR. Few questions

1. Does this somehow impact react hot loading? You could tests the impact of that after rebasing on master and checking this repo
2. Although this setup is nicer (so i'll might still merge this MR if RHL doesn't see new problems), the real problem in your example / test is that you are generating components on the fly during render. Please avoid doing this as this will kill React's reconciliation mechanism; every time you render the same thing, you render it with a completely new and different (as far as React knows) component, so React now has to render the entire subtree of components, as it never saw this type of component before. 

So instead of writing:

```
        function getWraped() {
            return inject(stores => ({
                name: stores.userStore.name
            }))(observer(NameDisplayer))
        }
```

You could simply write:

```
const WrappedNameDisplayer =  inject(stores => ({
                name: stores.userStore.name
            }))(observer(NameDisplayer))
        }
```

There is no need to define a new component here every time you are rendering the display name

zhangciwu:
For Q1, tested with mobx-react-boilerplate, it's all OK when I edited components render or store functions.

For Q2, the below code is like react pure function component, and can be more dynamic, like adding some dynamic options when generates this component. And I use this function for (nearly) root node, should be fine cause when store changes, root would not be re-rendered, only children nodes can be re-rendered

src/observer.js
```diff
@@ -367,16 +367,17 @@ export function observer(arg1, arg2) {
         throw new Error("Please pass a valid component to 'observer'")
     }
 
-    const target = componentClass.prototype || componentClass
+    const targetClass = class targetClassName extends componentClass {}
```
mweststrate: Wondering, will `targetClassName` now be the reported name in exceptions and React devtools and such?

src/observer.js
```diff
@@ -367,16 +367,17 @@ export function observer(arg1, arg2) {
...
+    const target = targetClass.prototype || targetClass
```
mweststrate: Is the right branch ever taken?

mweststrate:
Ok, thanks for the clarification! Left some more questions. Might be nice to split the code flow a bit to make explicitly clear which code is taken for a component class / function component / class react.createClass component, etc?

src/observer.js
ematipico: It will, so it's worth implementing `.displayName`

zhangciwu: Yeah, just added `displayName` for `targetClass`

zhangciwu:
Actually there is code to tell if it's `Stateless function component` at line 342 and above, and handling the branch after judging.

For `component class` / `class react.createClass`, just extend should be fine, cause `extend` is actually dealing with prototype of parent class or class-like-function. 
This extend just put `componentClass.prototype` into `targetClass.prototype.prototype` (to prevent `componentClass.prototype` being modified), should works fine for both `component class` / `class react.createClass`

mweststrate:
Ok, thanks! Will test this extensively and if everything works out fine
make part of the next release

Op ma 20 aug. 2018 om 12:47 schreef zhangciwu <notifications@github.com>:

mweststrate:
@Strate I think that should keep working as is because the new component inherits from the old one? Probably we should make sure we have some unit tests that make sure we don't have regressions here

zhangciwu:
Inheritance wouldn't drop class methods, including static ones, just seek one more step on prototype chain. (So it's called inheritance)
Unit tests are welcome to test corner cases, @Strate if you have some code which is broken after this, just post it here.

Strate:
@zhangciwu @mweststrate oh, sorry, I've missed thing that newely created class extends original.

mweststrate:
No problem! I'm happy you critically reviewed the code :) So feel free to
leave any concerns!

Op di 18 sep. 2018 om 08:56 schreef Artur Eshenbrener <
notifications@github.com>:

mweststrate:
When trying to use the changes locally, I get the following error on any class component:

```javascript
Uncaught TypeError: Class constructor Router cannot be invoked without 'new'
    at new targetClassName (:8001/static/builds/7.15.1/main.js:1183)
    at constructClassInstance (:8001/static/builds/7.15.1/vendor.js:221916)
    at updateClassComponent (:8001/static/builds/7.15.1/vendor.js:223400)
    at beginWork (:8001/static/builds/7.15.1/vendor.js:223786)
    at performUnitOfWork (:8001/static/builds/7.15.1/vendor.js:225785)
    at workLoop (:8001/static/builds/7.15.1/vendor.js:225849)
    at HTMLUnknownElement.callCallback (:8001/static/builds/7.15.1/vendor.js:216103)
    at Object.invokeGuardedCallbackDev (:8001/static/builds/7.15.1/vendor.js:216142)
    at invokeGuardedCallback (:8001/static/builds/7.15.1/vendor.js:215999)
    at renderRoot (:8001/static/builds/7.15.1/vendor.js:225927)
```

zhangciwu:
How do you import `observer`?
If you import from local `mobx-react` package, it would use `module` and `typings` entries in packages.json, the `module` entry is compiled file, would not preserve `class` keyword; `typings` entry is type declaration, just type definations.

If you import from '..somepath/mobx-react/src', `class` keyword would preserve, but this is not recommanded, try modified to the way above.

mweststrate:
`class` from mobx-react is not preserved, (using `import from "mobx-react"`), it is the `class` of Router that is being preserved and used here (since, targetting ES6)

zhangciwu:
Seems `Router`'s usage or rendering procedure may cause this (maybe), but not very clear.
What if use older version?

`at new targetClassName (:8001/static/builds/7.15.1/main.js:1183)`
This line and several lines around (of compiled main.js) are convinient to post?

mweststrate:
sure

```javascript
  var targetClass =
  /*#__PURE__*/
  function (_componentClass) {
    _inherits(targetClassName, _componentClass);

    function targetClassName() {
      _classCallCheck(this, targetClassName);

      return _possibleConstructorReturn(this, _getPrototypeOf(targetClassName).apply(this, arguments));
    }

    return targetClassName;
  }(componentClass);

  targetClass.displayName = componentClass.displayName || componentClass.name;
  var target = targetClass.prototype || targetClass;
  mixinLifecycleEvents(target);
  targetClass.isMobXReactObserver = true;
  makeObservableProp(target, "props");
  makeObservableProp(target, "state");
  var baseRender = target.render;
```

zhangciwu:
https://stackoverflow.com/questions/36577683/babel-error-class-constructor-foo-cannot-be-invoked-without-new

This shall be the root cause
you cannot extend a native class with a transpiled class.

Thus, you still need to compile source code to es5 or compile mobxreact into es6

mweststrate:
Both are not options imho:
* we cannot force all mobx-react consumers to transpile down their dependencies from es6 to es5
* we cannot force all mobx-react consumers to _not_ target es6

zhangciwu:
If need to publish es2015 (preserve class)
This might help

https://babeljs.io/blog/2018/06/26/on-consuming-and-publishing-es2015+-packages

    A common suggestion is for libraries to start publishing ES2015 under another field like es2015, e.g. "es2015": "es2015/package.mjs".
```
// @angular/core package.json
{
  "main": "./bundles/core.umd.js",
  "module": "./fesm5/core.js",
  "es2015": "./fesm2015/core.js",
  "esm5": "./esm5/core.js",
  "esm2015": "./esm2015/core.js",
  "fesm5": "./fesm5/core.js",
  "fesm2015": "./fesm2015/core.js",
}
```

zhangciwu:
As I noticed, mobx 5.x is ES6 only, 4.x is ES5 compatible
Maybe open a new major version to preserve class in this major version, and maintain two major versions like mobx?

sergei-startsev:
Here's a related issue in babel: https://github.com/babel/babel/issues/9367.

PR #562, opened 2018-09-28
Fix windows compilation + remove nscript due to fibers dep

xaviergonz:
Basically the same that was done for mobx in https://github.com/mobxjs/mobx/pull/1735

This PR:

* removes nscript due to a dependency with fibers, a native node module that was kinda hard to compile (and actually impossible with new node versions due to compilation errors)
* fixes the publish build script so it doesn't use such dependency and uses shelljs instead
* fixes publish and build script so they work in windows as well

I think this should help new contributors get up and running faster

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "test": "jest && npm run test:ts",
     "test:ts": "tsc -p test/ts",
     "test:travis": "npm run build && jest && npm run test:ts",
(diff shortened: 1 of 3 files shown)


xaviergonz:
gonna merge since build passes and it is basically the very same thing that was already merged in the pr for mobx

mweststrate:
No prob! Thanks!

PR #563, opened 2018-09-28
disposeOnUmount + new patch lifecycle method + updated deps

### LINKED ISSUE #561: onWillUnmount decorator to auto dispose reactions
a decorator that make disposers run on componentWillUnmount as seen in https://github.com/mobxjs/mobx/issues/1698

```
import {onWillUnmount, observer} from "mobx-react"

@observer
class Test extends React.Component {
   @onWillUnmount
   fetchData = reaction(....) // would also work for autorun, when, etc...

    // non decorator version?
    fetchData2 = onWillUnmount(this, reaction(.....))
}
```

xaviergonz:
This pr adds a disposeOnUnmount as seen in https://github.com/mobxjs/mobx-react/issues/561

Closes #561

Also since I was getting weird interactions between `@observer` and this decorator (since both try to tweak componentWillUnmount) I had to adjust the way of patching lifecycle methods. The good thing is that I think it is more robust (should work with native classes, works with lifecycle methods set as arrow functions, is able to compose "mixins", etc)

Also updated dependencies

### DIFF
--- a/src/disposeOnUnmount.js
+++ b/src/disposeOnUnmount.js
@@ -4,7 +4,7 @@ import { patch } from "./utils/utils"
 const storeKey = "__$mobxDisposeOnUnmount"
 
 function customComponentWillUnmount() {
-    for (const propKeyOrFunction of this[storeKey]) {
+    this[storeKey].forEach(propKeyOrFunction => {
         const prop =
             typeof propKeyOrFunction === "string" ? this[propKeyOrFunction] : propKeyOrFunction
         if (prop !== undefined && prop !== null) {
@@ -15,7 +15,7 @@ function customComponentWillUnmount() {
             }
             prop()
         }
-    }
+    })
     this[storeKey] = []
 }
 
--- a/src/disposeOnUnmount.js
+++ b/src/disposeOnUnmount.js
@@ -1,9 +1,9 @@
 import * as React from "react"
-import { patch } from "./utils/utils"
+import { patch, newSymbol } from "./utils/utils"
 
-const storeKey = "__$mobxDisposeOnUnmount"
+const storeKey = newSymbol("DisposeOnUnmount")
 
-function customComponentWillUnmount() {
+function runDisposersOnWillUnmount() {
     this[storeKey].forEach(propKeyOrFunction => {
         const prop =
             typeof propKeyOrFunction === "string" ? this[propKeyOrFunction] : propKeyOrFunction
@@ -42,7 +42,7 @@ export function disposeOnUnmount(target, propertyKeyOrFunction) {
 
     // tweak the component class componentWillUnmount if not done already
     if (!componentWasAlreadyModified) {
-        patch(target, "componentWillUnmount", customComponentWillUnmount, false)
+        patch(target, "componentWillUnmount", runDisposersOnWillUnmount, false)
     }
 
     // return the disposer as is if invoked as a non decorator
--- a/src/disposeOnUnmount.js
+++ b/src/disposeOnUnmount.js
@@ -1,0 +1,63 @@
+import * as React from "react"
+
+const storeKey = "__$mobxDisposeOnUnmount"
+
+function customComponentWillUnmount() {
+    for (const propKeyOrFunction of this[storeKey]) {
+        const prop =
+            typeof propKeyOrFunction === "string" ? this[propKeyOrFunction] : propKeyOrFunction
+        if (prop !== undefined && prop !== null) {
+            if (typeof prop !== "function") {
+                throw new Error(
+                    "[mobx-react] disposeOnUnmount only works on functions such as disposers returned by reactions, autorun, etc."
+                )
+            }
+            prop()
+        }
+    }
+    this[storeKey] = []
+}
+
+export function disposeOnUnmount(target, propertyKeyOrFunction) {
+    if (!target instanceof React.Component) {
+        throw new Error("[mobx-react] disposeOnUnmount only works on class based React components.")
+    }
+
+    if (typeof propertyKeyOrFunction !== "string" && typeof propertyKeyOrFunction !== "function") {
+        throw new Error(
+            "[mobx-react] disposeOnUnmount only works if the parameter is either a property key or a function."
+        )
+    }
+
+    // add property key / function we want run (disposed) to the store
+    const componentWasAlreadyModified = !!target[storeKey]
+    target[storeKey] = target[storeKey] || []
+    const store = target[storeKey]
+
+    store.push(propertyKeyOrFunction)
+
+    // tweak the component class componentWillUnmount if not done already
+    if (!componentWasAlreadyModified) {
+        let oldComponentWillUnmount = target.componentWillUnmount
+
+        Object.defineProperty(target, "componentWillUnmount", {
+            get: () => {
+                return function(...args) {
+                    if (oldComponentWillUnmount) {
+                        oldComponentWillUnmount.apply(this, args)
+                    }
+
+                    customComponentWillUnmount.call(this)
+                }
+            },
+            set: value => {
+                oldComponentWillUnmount = value
+            }
+        })
+    }
+
+    // return the disposer as is if invoked as a non decorator
+    if (typeof propertyKeyOrFunction !== "string") {
+        return propertyKeyOrFunction
+    }
+}
--- a/src/disposeOnUnmount.js
+++ b/src/disposeOnUnmount.js
@@ -20,6 +20,10 @@ function customComponentWillUnmount() {
 }
 
 export function disposeOnUnmount(target, propertyKeyOrFunction) {
+    if (Array.isArray(propertyKeyOrFunction)) {
+        return propertyKeyOrFunction.map(fn => disposeOnUnmount(target, fn))
+    }
+
     if (!target instanceof React.Component) {
         throw new Error("[mobx-react] disposeOnUnmount only works on class based React components.")
     }
--- a/src/disposeOnUnmount.js
+++ b/src/disposeOnUnmount.js
@@ -1,4 +1,5 @@
 import * as React from "react"
+import { patch } from "./utils/utils"
 
 const storeKey = "__$mobxDisposeOnUnmount"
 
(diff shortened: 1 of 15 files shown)


mweststrate:
I'm on holiday coming weeks (without laptop) so there will be some delay in reviewing, but I'll do my best to catch up the week after :-)

xaviergonz:
No problem :) Just updated dependencies as well.
also now travis shows the same console.error I get while running it in my pc (since I switched it to use yarn instead of npm)

xaviergonz:
fixed it, it was "for (const x of a)" transpiling into using Symbol.iterator, so just changed them to foreach

src/disposeOnUnmount.js
```diff
@@ -0,0 +1,53 @@
...
+
+    if (typeof propertyKeyOrFunction !== "string" && typeof propertyKeyOrFunction !== "function") {
+        throw new Error(
+            "[mobx-react] disposeOnUnmount only works if the parameter is either a property key or a function."
+        )
+    }
+
+    // add property key / function we want run (disposed) to the store
+    const componentWasAlreadyModified = !!target[storeKey]
+    target[storeKey] = target[storeKey] || []
```
mweststrate: NIT: simplify these 2 lines to: `const store = target[storeKey] || (target[storeKey] = [])`

src/index.d.ts
```diff
@@ -58,6 +58,15 @@ export function inject<S, P, I, C>(
 // ):
 // <T extends IReactComponent<P & S>(target: T) => IReactComponent<P> & IWrappedComponent<T>
 
+/**
+ * disposeOnUnmount
+ */
+export function disposeOnUnmount(target: React.Component, propertyKey: string): void
```
mweststrate: The official type for a field decorator is: `(target: Object, property: PropertyKey, descriptor?: PropertyDescriptor): void`

src/index.d.ts
```diff
@@ -58,6 +58,15 @@ export function inject<S, P, I, C>(
...
+export function disposeOnUnmount<TF extends Function | Function[]>(
```
mweststrate: I would explicitly type `Function` as `Disposer`, where `Disposer` is `() => void`

src/index.js
```diff
@@ -21,6 +21,7 @@ export {
 
 export { default as Provider } from "./Provider"
 export { default as inject } from "./inject"
+export { disposeOnUnmount } from "./disposeOnUnmount"
```
mweststrate: For naming, the name is clear, but quite long. Would there be a downside to calling it `onUnmount`?

mweststrate: Not entirely sure though

src/utils/utils.js
```diff
@@ -3,3 +3,59 @@ export function isStateless(component) {
...
+
+function getMixins(target, methodName) {
+    const mixins = (target["__$mobxMixins"] = target["__$mobxMixins"] || {})
+    const methodMixins = (mixins[methodName] = mixins[methodName] || {})
+    methodMixins.pre = methodMixins.pre || []
+    methodMixins.post = methodMixins.post || []
+    return methodMixins
+}
+
+export function patch(target, methodName, mixinMethod, runMixinFirst = false) {
```
mweststrate: could you elaborate why this enhanced patch was needed?

mweststrate:
@xaviergonz thanks! Great PR :-). Left some questions

src/utils/utils.js
xaviergonz: for two reasons. 
the first one is that the patching method would only support one decorator override, so when this decorator was being used with observer one of the two would run, but not both.
the second one is that the current patch method works OK for prototype methods, but not for property methods defined as arrow function, so this new patch method supports both

src/index.js
xaviergonz: I'm open for suggestions, so up to you 😁

xaviergonz: I'm open for suggestions, so up to you 😁

src/index.d.ts
xaviergonz: will do

src/index.d.ts
xaviergonz: I can change the others, but should the target be changed? as it is right now it prevents the decorator from being used on non react component classes

src/disposeOnUnmount.js
xaviergonz: will do

src/index.js
xaviergonz: I'm just weary of how it'd look on the non decorator version
onUnmount(reaction...)
makes it sound like the reaction will trigger on unmount or something

src/index.d.ts
xaviergonz: Just took a look at lib.es5.d.ts and it shows
`declare type PropertyDecorator = (target: Object, propertyKey: string | symbol) => void;`
since the decorator should only work over react component classes and it doesn't work over symbol properties, shouldn't I leave it as is?

xaviergonz:
code review changes done, except for:
* the decorator typing, see comments I added about it
* name still not changed, (see comment I added about it)

src/utils/utils.js
xaviergonz: Quoting the PR description:
Also since I was getting weird interactions between @observer and this decorator (since both try to tweak componentWillUnmount) I had to adjust the way of patching lifecycle methods. The good thing is that I think it is more robust (should work with native classes, works with lifecycle methods set as arrow functions, is able to compose "mixins", etc)

src/index.js
mweststrate: Ok, let's keep it as is! If it is to long people can always alias during import :)

xaviergonz:
so is all good now?

mweststrate:
Will try to review the rest later tonight!

Op do 11 okt. 2018 om 20:50 schreef Javier Gonzalez <
notifications@github.com>:

mweststrate:
(currently at meetup)

Op do 11 okt. 2018 om 21:03 schreef Michel Weststrate <mweststrate@gmail.com

xaviergonz:
no need to hurry :) Btw, I just created a require('@mweststrate') tag on MST for issues that cannot advance until you make a decision, just so you can quickly filter over issues that absolutely require you (eventually) :)

src/disposeOnUnmount.js
```diff
@@ -0,0 +1,52 @@
+import * as React from "react"
+import { patch } from "./utils/utils"
+
+const storeKey = "__$mobxDisposeOnUnmount"
+
+function customComponentWillUnmount() {
+    this[storeKey].forEach(propKeyOrFunction => {
```
mweststrate: NIT: `this[storeKey].splice(0).forEach` prevents concurrent modifications, and doesn't need the reset of line 19. (No need to change it if there are no other changes for this MR, just a nice pattern in general, imho)

src/disposeOnUnmount.js
```diff
@@ -0,0 +1,52 @@
+import * as React from "react"
+import { patch } from "./utils/utils"
+
+const storeKey = "__$mobxDisposeOnUnmount"
+
+function customComponentWillUnmount() {
```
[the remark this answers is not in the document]
xaviergonz: sounds good too :)

src/disposeOnUnmount.js
xaviergonz: OK, actually forEach does, but (for of) doesn't :)
still, since this once uses forEach it should be ok

src/utils/utils.js
```diff
@@ -3,3 +3,59 @@ export function isStateless(component) {
...
+            realMethod.apply(this, args)
+        }
+
+        mixins.post.forEach(post => {
+            post.apply(this, args)
+        })
+    }
+    getFunction["__$mobxMixin"] = true
+
+    const newDefinition = {
```
mweststrate: Optimization: performance / memory consumption is significant lower if property descriptors are cached and reused per name. To achieve that `getFunction` needs to be generic, for that, `realMethod` needs not to be used. That could be done by for example storing the real method on the `mixins.real`  property

src/utils/utils.js
```diff
@@ -3,3 +3,59 @@ export function isStateless(component) {
...
+        get: () => getFunction,
+        set: value => {
+            realMethod = value
+        },
+        configurable: false
```
mweststrate: keep this one on `true`. Many testing tools (if people want to stub) or things like HMR trip over non-configurable properties

test/ts/compile-ts.tsx
```diff
@@ -232,14 +270,34 @@ class InjectSomeStores extends Component<{ x: any }, {}> {
...
-		render() {
-			return <div/>
-		}
-	}
-
-	const Injected = inject("test")(T);
-	<Injected.wrappedComponent x={3} />
-}
\ No newline at end of file
+    // just to make sure it compiles
```
mweststrate: It would be great if we had an actual Typescript test for this, as the decorator implementation of TS is different from the babel one, but it might be too much of a hassle to add it to the current setup, I am confident enough that this case is simple enough to not fail (famous last words)

src/utils/utils.js
```diff
@@ -3,3 +3,59 @@ export function isStateless(component) {
     // `() => {}` via Babel has prototype too.
     return !(component.prototype && component.prototype.render)
 }
+
+function getMixins(target, methodName) {
+    const mixins = (target["__$mobxMixins"] = target["__$mobxMixins"] || {})
```
mweststrate: NIT: constants (and Symbols)

src/utils/utils.js
xaviergonz: ok, hopefully those tools won't break it badly :)

src/utils/utils.js
xaviergonz: sure

mweststrate:
@xaviergonz Ok, I think I finished my review :-). Only reservation is that I think double-patching the old way should work fine, and has less risk of conflicting with other tools downstream. On the other hand, fixing the subtle arrow func issue nice, as people run into that a few times. So let's proceed with this approach unless proven otherwise. I think it deserves to be a minor for that release at least though :-).

xaviergonz:
code review changes done, while at it i also changed some other internal props to symbols

xaviergonz:
ah forgot the Optimization: performance / memory consumption one

xaviergonz:
definition optimization done as well

xaviergonz:
There's one thing that worries me about the definition optimization though.
If two different classes use the same prop def for the same method and a tool decides only to patch up one of the components, will it end up patching both?
If so, should the definition be cached per class as well?

mweststrate:
@xaviergonz if a tools patches something, the descriptor gets replaced, so the descriptors themselves are practically immutable, making this save (afaik grabbing a property descriptor using reflection, and then changing it has no effect)

xaviergonz:
while at it also improved a bit the symbol polyfill (it was based on math.random, which seemed flaky) and unified them all over the utils function newSymbol

mweststrate:
Ok, will try to merge tomorrow, some things I want to check locally, so hope I'll find tomorrow time to wrap up :)

PR #573, opened 2018-10-16
neglectable to negligible

ThaJay:
https://english.stackexchange.com/questions/202832/is-there-a-difference-between-negligible-and-neglectable


mattruby:
Thanks!

PR #578, opened 2018-10-18
Fix conditional disposeOnUnmount, also no pesky DS_Store files

JefHellemans:


mweststrate:
Released as 5.3.3

xaviergonz:
@JefHellemans Just wondering, what does this patch fix exactly?
In theory a precondition for patch(runDisposersOnWillUnmount) is that this[storeKey] is present

JefHellemans:
@xaviergonz the implementation patches `componentWillUnmount` on prototype level, not instance level. This means that if you use `disposeOnUnmount` conditionally (like we are doing in some components), the prototype still tries to run `forEach` on `this[storeKey]` while some instances don't have it.

Maybe this should be fixed in the `patch` implementation actually, but that's a different story.

Sorry for the late reply as well by the way 😊

PR #582, opened 2018-10-19
fix recursive calls for patching

### LINKED ISSUE #579: Maximum call stack size exceeded error in lifecycle methods
I encountered this error when upgrading to v5.3.2.  (It doesn't happen in 5.2.x.)  The issue is that certain lifecycle methods end up recursing infinitely—the callstack ends up looking something like this:

```
RangeError: Maximum call stack size exceeded
    at c.wrapperMethod
    at c.f.componentDidMount
    at c.wrapperMethod
    at c.f.componentDidMount
    at c.wrapperMethod
    at c.f.componentDidMount
    at c.wrapperMethod
    at c.f.componentDidMount
    at c.wrapperMethod
    at c.f.componentDidMount
    ....
```

It appears this happens when `@observer` patches a component's lifecycle methods *and* something else is patching those methods as well, such as another decorator or a base class.  (For instance [the BaseComponent in office-ui-fabric-react](https://github.com/OfficeDev/office-ui-fabric-react/blob/1b9405b3b64839975e16ee1fad04d7868d9e3b99/packages/utilities/src/BaseComponent.ts) does this.)  I haven't dug into the mobx-react code enough to grok exactly what the problem is, but I've seen multiple instances of mobx-react interacting badly with other libraries that do this sort of thing.

xaviergonz:
Fixes #579 by making sure recursive calls to the function are stopped

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -48,6 +48,7 @@ function createOrGetCachedDefinition(methodName, enumerable) {
         mixins.realRunning = true
 
         const realMethod = mixins.real
+        let retVal
 
         try {
             mixins.pre.forEach(pre => {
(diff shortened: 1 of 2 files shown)


mweststrate:
Released as 5.3.4

PR #590, opened 2018-10-27
observableProps implementation

xaviergonz:
### observableProps(componentInstance) _[Requires MobX >= 5]_

Function that converts your props into fully reactive (observable) props.
This allows you to, for example, create reactions/autorun/computed over individual props rather than the whole props object.

```javascript
import { observer, observableProps } from "mobx-react"

<NameAndAge first="Mary" last="Poppins" age={30} />

@observer
class NameAndAge extends React.Component {
    obsProps = observableProps(this)

    @computed
    get fullName() {
        // this computed will only recompute when any of these props change
        // if we were using props instead it would also recompute if age
        // (or any other property) changed
        return this.obsProps.first + " " + this.obsProps.last;
    }
}
```

A non breaking change since it is opt-in
requires mobx5 since it requires the proxy to detect when props is adding a new property, will throw if that capability is not present (this is, mobx <= 4)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,18 +5,14 @@
     "version": "0.2.0",
     "configurations": [
         {
-            // Note; this config requires node 8.4 or higher
             "type": "node",
-            "protocol": "auto",
+            "name": "debug unit tests",
             "request": "launch",
-            "name": "debug unit test",
-            "stopOnEntry": false,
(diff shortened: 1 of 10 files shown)


xaviergonz:
basically it is a reactive-props in an opt-in fashion, and requires the user to agree on the new behaviour 

@mweststrate thoughts?

PR #595, opened 2018-11-01
withObservableProps

xaviergonz:
### withObservableProps(component, options?: { deepProps: boolean | string[] })

_[Requires MobX >= 5]_

High order components that converts your props into fully reactive (observable) props.

This allows you to, for example, create reactions/autorun/computed over individual props rather than the whole props object.
Using observable props also means it will make your wrapped component use an optimized re-render
mode where the component will only re-render when any of the actual properties used by the last render have changed, rather than any time they change.

**The only requirement to use is feature is that your wrapped component must access the properties from `this.props.obs` rather than `this.props`.**

By default individual props are turned into shallow observables (unless they were already observables in themselves, then they are unaffected),
but this can be changed with the `deepProps` option:

-   `false` (default): all properties are turned into shallow observables.
-   `true`: all properties are turned into deep observables (plain objects, arrays and maps are supported).
-   `string[]`: array with the name of the properties to make deep observables, all others will be kept shallow observables.

Example (Javascript):

```js
import { computed } from "mobx"
import { withObservableProps } from "mobx-react"

const NameAndAge = withObservableProps(
    class extends React.Component {
        @computed
        get fullName() {
            // this computed will only recompute when any of these props change
            // if we were using props instead it would also recompute if age
            // (or any other property) changed
            // not how this.props.obs is used rather than this.props
            return this.props.obs.first + " " + this.props.obs.last;
        }

        render() {
            // render will only trigger when either first or last properties change,
            // but not when age changes, since it is unused by the render method
            return (<div>${this.fullName}</div>);
        }
    }
)

// note how even though inside the wrapper component props must be
// accessed through this.props.obs, here they are just plain properties
<NameAndAge first="Mary" last="Poppins" age={30} />
```

Example (TypeScript):

```typescript
import { computed } from "mobx"
import { withObservableProps, ObservableProps } from "mobx-react"

interface NameAndAgeProps {
    first: string;
    last: string;
    age: number;
}

const NameAndAge = withObservableProps(
    class extends React.Component<ObservableProps<NameAndAgeProps>> {
        @computed
        get fullName() {
            // this computed will only recompute when any of these props change
            // if we were using props instead it would also recompute if age
            // (or any other property) changed
            // not how this.props.obs is used rather than this.props
            return this.props.obs.first + " " + this.props.obs.last;
        }

        render() {
            // render will only trigger when either first or last properties change,
            // but not when age changes, since it is unused by the render method
            return (<div>${this.fullName}</div>);
        }
    }
)

// note how even though inside the wrapper component props must be
// accessed through this.props.obs, here they are just plain properties
<NameAndAge first="Mary" last="Poppins" age={30} />
```


xaviergonz:
I tried making it work using this.props directly rather than this.props.obs, but react will just break (well, warn actually) when props passed to super() is not the same than the one passed on the constructor

mweststrate:
Hey @xaviergonz, 

the idea is clear, but I am a bit hesitant to add even more ways to mobx-react to work with props etc, I think mobx is already mingling to much in that area. I think in the cases where this is a problem, in most cases it can be fixed by setting the `equals` method of the `@computed` to `comparer.shallow` or by lifting out the computed to a model class. 

The `.obs.` looks quite confusing from api perspective and not very self explanatory.

xaviergonz:
Well, computed with comparer shallow wouldn't be good enough, since (given the example above) any property change, even if not used (in this case age) would still trigger an unneeded re-render.

Using a model object with all observables does indeed work, but it makes it:
a) kind of ugly to use (you'd write something like `<C allProps={someObject}/>`)
b) forces the component user to ensure that the properties are observable (in this case the properties are turned into observables by the component itself)

To be honest, one of the main use cases I have for this is for style, where using this:
`<C style={{ margin: 5 }}/>`
always make the child component re-render, no matter what (even if it is observable)
by using this "trick" it won't actually re-render, it will just "deep assign" the object into the same observable object (assuming deepProps: ['style'] is used)

But I can see how this might be too specific to be into mobx-react, no worries, I can just turn it into an npm library :)

That being said, would the reconciliation algorithm (the deep assign of an standard object into an observable object) be useful for mobx-utils perhaps?

This is, given an observable object like
`
const myObs = observable({ x: 5, y: {}, z: 1 })
`
and a non observable object like
`const nonObs = { x: 5, y: { a: 6 }, z: 2}`

`observableDeepAssign(myObs, nonObs, { deep: true })` // in this case I guess true would be the default

be useful in mobx-utils?

xaviergonz:
Also, maybe this could be a good case the lite/hooks/functional version?
`
const C = observer((props) => {
  ...
}, { observableProps: true/false/{deepProps: true/false/array} });
`

mweststrate:
Added breaking change label, as MobX 5 would be required

mweststrate:
Closing for inactivity, and since this whole problem is taking a different direction since hooks (see also related disscusions in mobx-react-lite).

PR #603, opened 2018-11-12
Add link to mobx-react-lite

FredyC:


xaviergonz:
Well, future proof looks a bit exaggerated when the hooks themselves are in alpha :)

FredyC:
@xaviergonz I am open to a better wording :) I meant it more like that it won't be held hostage by some old tech :) Just good old bleeding edge tech :)

xaviergonz:
How about
"If you are using [React hooks](https://reactjs.org/docs/hooks-intro.html) with latest React 16.7 and like living on the bleeding edge then have a look at the new [mobx-react-lite](https://github.com/mobxjs/mobx-react-lite)."
?

Issue #602, opened 2018-11-09
makeComponentReactive throws "TypeError: Cannot read property 'displayName' of undefined"


dgcoffman:
When `observer` directly wraps a styled component from the very common library `styled-components`, the following error is thrown:
```
Uncaught TypeError: Cannot read property 'displayName' of undefined
    at makeComponentReactive (mobx-react/index.module.js:857)
    at target.render (mobx-react/index.module.js:1068)
    at updateForwardRef (react-dom/cjs/react-dom.development.js:8254)
    at beginWork (react-dom/cjs/react-dom.development.js:9002)
    at performUnitOfWork (react-dom/cjs/react-dom.development.js:11814)

The above error occurred in the <ForwardRef> component
```

Here's my code https://codesandbox.io/s/k3np5pvmm3
```
import * as React from 'react';
import { render } from 'react-dom';
import { observer } from 'mobx-react';
import styled from 'styled-components';

const A = styled.a`
  height: auto;
`;

const C = observer(A);

render(<C />, document.getElementById('root'));
```

package.json
```
"react": "16.6.1",
"styled-components": "4.0.3",
"mobx": "5.6.0",
"mobx-react": "5.3.6",
```

Here's the Chrome debugger on the line that throws
<img width="1666" alt="screen shot 2018-11-09 at 11 08 38 am" src="https://user-images.githubusercontent.com/1711204/48282920-e4daff80-e40f-11e8-882b-0ea78c4bf05f.png">

Apparently `this` is undefined in the upstream https://github.com/mobxjs/mobx-react/blob/2e2cf2241baaa3a47e8a125645dade57a2186927/src/observer.js#L356

This only occurs with styled-components 4+ -- likely because they have switched to using React.forwardRef.

You will likely need some forwardRef check like 
```
function isForwardRef(object) {
    return typeOf(object) === Symbol.for('react.forward_ref');
}
```

Relevant code from styled-components: https://github.com/styled-components/styled-components/blob/2975ff9f217c108737fa250bb3d8b38e017d9fa8/src/models/StyledComponent.js#L229

Any ideas?

Similar to https://github.com/mobxjs/mobx-react/issues/477

pimterry:
This is a more general problem: mobx-react does not support any components that use `React.forwardRef`.

The discussion of the same problem and corresponding solution in redux-connect will probably be useful: https://github.com/reduxjs/react-redux/issues/914.

mweststrate:
[...]

PR #606, opened 2018-11-17
Create a failing test for React.forwardRef

pimterry:
This is a failing test to illustrate https://github.com/mobxjs/mobx-react/issues/602. It fails with:

```
Error: Uncaught [TypeError: Cannot read property 'displayName' of undefined]
```

I'm pretty confident that this should work, and that this is the underlying issue that breaks styled components: any components that use [ref forwarding](https://reactjs.org/docs/forwarding-refs.html) can't be passed to `observe()`.


mweststrate:
Thanks! This was really  helpful :)

PR #618, opened 2018-12-05
feat: add version to built file (resolve #610)

### LINKED ISSUE #610: Version number at the top of the file
Would it be possible please to add the version number at the top of the file like react does, so we know which version we are looking at?  Something like https://i.imgur.com/cewiKjq.png

Thanks!

fi3ework:


mweststrate:
@fi3ework note that this won't suffice, in the publish flow the build is run first, and after that the version number is updated, so some swapping around is needed there as well :)

Issue #616, opened 2018-12-04
TypeScript, MobX-React, and RefObjects


42shadow42:
When using a combination of TypeScript, MobX-React, and RefObjects, it becomes impossible to use strong typed RefObjects:

private searchForm: RefObject<SearchForm> = React.createRef();

    @action
    onSearchSelected = (search: Search) => {
        this.searchForm.current.populateSearchForm(search); //This compiles but fails at runtime
        this.searchForm.current.wrappedInstance.populateSearchForm(search); //This succeeds at runtime, but won't compile
    }

I believe this can be resolved, by using the new [forwardRef API](https://reactjs.org/blog/2018/03/29/react-v-16-3.html) for react components

42shadow42:
@mweststrate I saw you fixed issues with reactivity and forwardRef on Observer, do you have any insight that might help fix reactivity of forwardRef on injector? I would really appreciate it!

42shadow42:
Fixed by #619

PR #619, opened 2018-12-05
TypeScript, MobX-React, and RefObjects

42shadow42:
Fixed issue where inject decorated classes were not compatible with strongly typed RefObjects

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -1,4 +1,4 @@
-import { Component, createElement } from "react"
+import React, { Component, createElement } from "react"
 import hoistStatics from "hoist-non-react-statics"
 import * as PropTypes from "./propTypes"
 import { observer } from "./observer"
@@ -113,7 +113,7 @@ export default function inject(/* fn(stores, nextProps) or ...storeNames */) {
                 // see #111
                 injected = observer(injected)
                 injected.isMobxInjector = true // restore warning
-                return injected
+                return createElement(injected, props)
             })
 
         }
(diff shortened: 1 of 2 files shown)


42shadow42:
Fixes #616 

This is missing unit tests currently and should be added, I'm not really an expert here so I wouldn't know how to write them myself.

This should be treated as a breaking change, as it affects the usage of wrappedInstance on RefObjects, the migration path is to remove the .wrappedInstance and use the item directly.

mweststrate:
@42shadow42 I need a little clarification. Is the goal to fix typings, or do you want to enable refForwarding on `inject`? Which of those two would be according to you the correct fix?

In either case, could you demonstrate what you are fixing in a small reproduction (codesandbox, unit test or otherwise), so that the solution can be verified?

42shadow42:
@mweststrate The goal is to fix typings, the proposed solution is refForwarding.

Here is my reproduction:
[mobx-injector.zip](https://github.com/mobxjs/mobx-react/files/2672680/mobx-injector.zip)

If you look at parent-component.tsx you will see that typescript enforces typings of child and child2 variables, but the actual runtime types are inject wrappedInstances so while the typescript compiles runtime fails, the proposed fix of refForwarding makes the compile time type match the runtime type.

mweststrate:
Thanks for the PR! I took the important changes, but refactored and cleaned up the as it got a bit messy in #644. Also added / updated tests. 

So closing this one in favor of that one, but thanks for pointing out the solution and will put your name on the contributors list!

PR #622, opened 2018-12-11
Cache newSymbol created Symbols

normano64:
The observer decorator created 2 Symbols for each decorated component
instead of reusing the Symbols for props and state that was previously
created

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -5,13 +5,21 @@ export function isStateless(component) {
 }
 
 let symbolId = 0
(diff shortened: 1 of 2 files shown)


mweststrate:
@normano64 looks valid. Did this solve any specific issue for you?

normano64:
@mweststrate I recently bumped mobx-react version in a project is running on an old version of JavaScriptCore caused our memory usage to go up by 56MB from all the Symbols. 😅

I should probably have mentioned in the MR that it was previously cached before commit

PR #632, opened 2019-01-21
Fix disposeOnUmount typo in README.md

WearyMonkey:


mweststrate:
Thanks for the catch!

PR #641, opened 2019-02-07
Decorator version of disposeOnUnmount now works with arrays (closes #637)

### LINKED ISSUE #637: disposeOnUnmount as decorator with arrays throw an error when unmounts
`"mobx-react": "^5.4.3",`

image
Then, when component unmounts:
>Uncaught Error: [mobx-react] disposeOnUnmount only works on functions such as disposers returned by reactions, autorun, etc.

`disposeOnUnmount(this, [...])` works perfectly.

Amareis:
#631 related

### DIFF
--- a/README.md
+++ b/README.md
@@ -432,6 +432,14 @@ class SomeComponent extends React.Component {
     // decorator version
     @disposeOnUnmount
     someReactionDisposer = reaction(...)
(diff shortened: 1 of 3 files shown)


mweststrate:
Released as 6.0.3

Issue #640, opened 2019-02-07
mobx-react 6


mweststrate:
Things to be addressed:

* [x] switch to modern context (but keep same Provider / inject api) (#448)
* [x] plan strategy for hook based components (currently; use `<Observer>` or `mobx-react-lite`)
* [x] ~make sure mobx-react is ready for concurrent react (primary concern: don't leak memory on early component disposal)~ -> opened #645 to address
* [x] dispose components on error? (since upstream error boundaries should now handle them)
* [x] merge #616 / #619 
* [x] stop lifecycle compatibility support
* [x] ~split into smaller packages for Provider / inject / propTypes?~ nope, see #661 
* [x] verify devtools behavior (especially functional comps)
* [x] verify component tree (names, provider & inject)
* [x] Drop dependency on `hoist-non-react-statics`?
* [x] rename entry files (#408)?
* [x] use microbundle?
* [x] update readme
* [x] process review comments
* [x] process TODO's

FredyC:
Hm, so you want to ditch the mobx-react-lite eventually? It probably should merge together at some point, but wouldn't it be too convoluted trying to support classes and hooks in one package?

FredyC:
Also, I convinced that we should teach people how to use React Context instead of including it in the package. For once it's super easy to do it and it doesn't seem it gives much of the benefit having it here. On the contrary, it would be harder to have proper typing for it.

mweststrate:
@FredyC definitely not, the question is more, merge it into mobx-react, or make it be the next mobx-react (dropping propTypes, context, provider / inject and such). But I doubt we can loose the latter on the short term, I think most people still prefer the mobx-react api over React's own context api. So, yeah I am basically having the same questions as you have :)

FredyC:
Switching mobx-react to a new context API would make a bit nicer and smaller repo code for sure. However, as long as people will be using `inject`, they won't really know there is a different tech internally.

Personally, I don't really see such a huge problem including one extra Provider in the tree to cover the transition. This isn't a first package suffering from that issue. But we should definitely make some migration guide toward explanation this gradual adoption.

urugator:
I would personally like to see prop-types/inject/etc in their own seperate packages.

FredyC:
[...]

PR #644, opened 2019-02-11
mobx-react v6

mweststrate:
See #640


src/index.js
```diff
@@ -9,6 +9,8 @@ if (!spy) throw new Error("mobx-react requires mobx to be available")
 if (typeof rdBatched === "function") configure({ reactionScheduler: rdBatched })
 else if (typeof rnBatched === "function") configure({ reactionScheduler: rnBatched })
 
+// TODO: re-export mobx-react-lite stuff?
```
FredyC: I think it should be safe enough as everything is prefixed with `use` so it should generally understood as a hook to be used in FC only.

src/observer.js
```diff
@@ -299,18 +301,12 @@ export function observer(componentClass) {
         !componentClass.isReactClass &&
         !Component.isPrototypeOf(componentClass)
     ) {
-        const observerComponent = observer(
```
FredyC: Nit: Update the comment on lines 293-295 to reflect this change.

src/observer.js
```diff
@@ -299,18 +301,12 @@ export function observer(componentClass) {
...
-                static contextTypes = componentClass.contextTypes
-                static propTypes = componentClass.propTypes
-                static defaultProps = componentClass.defaultProps
-                render() {
-                    return componentClass.call(this, this.props, this.context)
-                }
-            }
-        )
+        const observerComponent = observerLite(componentClass)
+        // TODO: move to mobx-react-lite
```
FredyC: I am fine with that. Regarding hoistStatics there is an open issue with some thoughts: https://github.com/mobxjs/mobx-react-lite/issues/32

src/observer.js
```diff
@@ -348,20 +344,13 @@ function mixinLifecycleEvents(target) {
     }
 }
 
-export const Observer = observer(({ children, inject: observerInject, render }) => {
+// TODO: use mobx-react-lite version?
```
FredyC: Should be safe enough imo since it is used in the component tree, but let's add a test to see what happens when wrapped component is class one.

mweststrate:
Besides processing a lot of TODO's, the biggest open question atm is how to integrate with the DevTools.

package.json
```diff
@@ -59,14 +59,14 @@
...
-    "rollup-plugin-node-resolve": "^3.0.0",
+    "rollup": "^1.1.2",
+    "rollup-plugin-alias": "^1.5.1",
+    "rollup-plugin-babel": "^4.3.2",
+    "rollup-plugin-commonjs": "^9.2.0",
+    "rollup-plugin-filesize": "^6.0.1",
+    "rollup-plugin-node-resolve": "^4.0.0",
     "rollup-plugin-replace": "^2.1.0",
-    "rollup-plugin-uglify": "^6.0.0",
+    "rollup-plugin-uglify": "^6.0.2",
```
FredyC: Perhaps you can replace uglify with actually maintained rollup-plugin-terser? Lite package is using that as well.

src/Provider.js
```diff
@@ -1,74 +1,47 @@
...
+import { Children, Component, createContext, createElement } from "react"
+import { shallowEqual } from "./utils/utils"
 
 const specialReactKeys = { children: true, key: true, ref: true }
 
-class Provider extends Component {
-    static contextTypes = {
-        mobxStores: PropTypes.objectOrObservableObject
-    }
+export const MobXProviderContext = createContext({})
```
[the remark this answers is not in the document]
mweststrate: No, want to keep `Provider / inject` as is. I think `React.createContext` is the way to go now. But I do want to keep supporting for `Provider / inject` at least in v6 to not make the migration harder. (would be a shame if people wouldn't migrate to hooks, because they lack the time to factor Provider out)

src/Provider.js
```diff
@@ -1,74 +1,47 @@
...
-        if (!prevState) return nextProps
-
-        // Maybe this warning is too aggressive?
-        if (
-            Object.keys(nextProps).filter(validStoreName).length !==
-            Object.keys(prevState).filter(validStoreName).length
-        )
-            console.warn(
+        const newStores = { ...prevState, ...grabStores(nextProps) } // spread in prevState for the context based stores
+        if (!shallowEqual(prevState, newStores))
```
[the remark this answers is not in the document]
mweststrate: Good one! Will do

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+* Changing the set of stores in `Provider` is no longer supported and while throw a hard error (this was a warning before), as the model of `Provider` / `inject` has always been to inject final values into the tree. (That is, fixed references, injected objects themselves can be stateful without problem). If you want to dynamically swap what is provided into the tree, use `React.createContext` instead of `Provider` / `inject`. The  suppressChangedStoreWarning` flag for `Provider` has been dropped.
+* The third argument of custom `storesToProps` functions passed to `inject` is no longer available.
+* `<Observer>` no longer supports the deprecated `inject` property.
+* Defining `shouldComponentUpdate` on `observer` compon
+* `propTypes` is no longer exposed, use `PropTypes` instead
+* `disposeOnUnmount` now only supports direct subclasses of `React.Component` and `React.PureComponent`. This prevents several unreliable edge cases that silently leaked memory before. Either only extend React.(Pure)Component when using `disposeOnUnmount`, or manually clean up stuff in `componentWillUnmount`.
```
mweststrate: @xaviergonz how you feel about this one?

package.json
mweststrate: Migrated to microbundle, so that abstracts all that away, significantly simplifying the setup :).

CHANGELOG.md
xaviergonz: fine by me. will it still work when arrow functions are used to declare lifecycles though?

CHANGELOG.md
```diff
@@ -1,5 +1,29 @@
...
+* `observer` components no longer automatically recover from errors (to prevent potential memory leaks). Instead, this is the responsibility of error boundaries.
+* `inject` now supports ref forwarding. As such, the `.wrappedInstance` property has been removed since refs can be used instead. (Fixes [#616](https://github.com/mobxjs/mobx-react/issues/616) (See also [#619](https://github.com/mobxjs/mobx-react/pull/619) by [42shadow42](https://github.com/42shadow42))
+* Changing the set of stores in `Provider` is no longer supported and while throw a hard error (this was a warning before), as the model of `Provider` / `inject` has always been to inject final values into the tree. (That is, fixed references, injected objects themselves can be stateful without problem). If you want to dynamically swap what is provided into the tree, use `React.createContext` instead of `Provider` / `inject`. The  suppressChangedStoreWarning` flag for `Provider` has been dropped.
+* The third argument of custom `storesToProps` functions passed to `inject` is no longer available.
+* `<Observer>` no longer supports the deprecated `inject` property.
+* Defining `shouldComponentUpdate` on `observer` compon
```
xaviergonz: this line is cut

src/observer.js
mweststrate: Done

src/observer.js
mweststrate: done

CHANGELOG.md
mweststrate: IIRC yes

CHANGELOG.md
mweststrate: Fixed

mweststrate:
mobx-react@6.0.0-rc.2 is now available!

samcooke98:
> mobx-react@6.0.0-rc.2 is now available!
Just a heads-up, it seems that it breaks when attempting to use `observer(someClassComponent)` 

https://codesandbox.io/s/j3k4p2jkv5

I did some digging, it seems `h` is the built `<Observer>` from `mobx-react-lite` - Not sure why it's not included in the bundle however?

FredyC:
The `h` is not the Observer, it should be `React.createElement`, but it's indeed missing in there for some reason. The `Observer` is minified under `y` and is there available.

mweststrate:
Fixed in mobx-react@6.0.0-rc.4

xaviergonz:
Are you using terser rather than uglify es? Terser fixed some react minification bugs

mweststrate:
No, jsx wasn't configured for microbundle, causing it to default to `h`
instead of `React.createElement`

FredyC:
So I think you should remove reexport of `useDisposable`, `useComputed` and `useObservable` as [I came to the conclusion](https://github.com/mobxjs/mobx-react-lite/issues/94) these should be removed from lite package anyway. Let's not confuse people more by exposing these here.

src/index.js
```diff
@@ -1,46 +1,17 @@
...
-    const mobx = { spy, extras: { getDebugName } }
-    const mobxReact = {
-        renderReporter,
-        componentByNodeRegistry,
-        componentByNodeRegistery: componentByNodeRegistry,
-        trackComponents
-    }
-    __MOBX_DEVTOOLS_GLOBAL_HOOK__.injectMobxReact(mobxReact, mobx)
-}
+export { Provider } from "./Provider"
```
FredyC: Note here we should probably export the Context object as well so people can make a custom hook for it. It's definitely better an easier migration than wanting from them to rewrite to new Context in a big swoop.

README.md
```diff
@@ -450,6 +409,17 @@ class SomeComponent extends React.Component {
 }
 ```
 
+## DevTools
+
+`mobx-react@6` and higher are no longer compatible with the mobx-react-devtools. 
+That is, the MobX react devtools will no longer show render timings or dependency trees of the component. 
+The reason is that the standard React devtools are no also capable of highlighting re-rendering components.
```
macbem: `devtools are no also capable` - what did you mean here? Looks like a typo.

FredyC: @macbern Use the suggestions next time please
```suggestion
The reason is that the standard React devtools are also capable of highlighting re-rendering components.
```

PR #671, opened 2019-04-24
Separates decorator/instance usage of disposeOnUnmount

### LINKED ISSUE #666: Mixing `@disposeOnUnmount` and `disposeOnUnmount(this, ...)` is broken
Hello,

`disposeOnUnmount` registers the disposers or the property keys (pointing to the disposers) in an array that is stored on the class. If you use `disposeOnUnmount` as a decorator, that array will be stored in the class prototype, but if you use it inside the class it will be stored as an instance property. `runDisposersOnUnmount` doesn't care where the property is located because `this[storeKey]` can refer to either.

As long as I have a `@disposeOnUnmount` decorator, the array will necessarily be stored on the prototype, but any `disposeOnUnmount` that is declared inside the class will also be registered on the prototype, despite being firmly from the instance side of things.

Usually, this doesn't cause a lot of problems, except for the case where you use two (or more) instances of the same class, with one unmounting before the other : the first unmount will also dispose every reaction created inside the instance of the second class.

I know that `disposeOnUnmount` doesn't distinguish between being called as a decorator and being called directly, but I feel that storing the property key disposers and the function disposers in two separate arrays could be an easy fix that doesn't break the API. I don't think people are using the `disposeOnUnmount(this, "myDisposer")` way, since the docs don't mention it and it's weirder and not as convenient as `disposeOnUnmount(this, autorun(() => …))`. Using the argument type as a way to separate decorator and non-decorator usage looks like an okay heuristic we could use, and in any case doing this won't break anything, so that would be a straight improvement.

JabX:
Fixes #666 by splitting the disposer store into two separate prototype and instance stores, depending on the target parameter.

### DIFF
--- a/src/disposeOnUnmount.js
+++ b/src/disposeOnUnmount.js
@@ -1,14 +1,11 @@
 import * as React from "react"
 import { patch, newSymbol } from "./utils/utils"
 
-const storeKey = newSymbol("disposeOnUnmount")
+const protoStoreKey = newSymbol("disposeOnUnmountProto")
+const instStoreKey = newSymbol("disposeOnUnmountInst")
 
 function runDisposersOnWillUnmount() {
-    if (!this[storeKey]) {
-        // when disposeOnUnmount is only set to some instances of a component it will still patch the prototype
-        return
-    }
-    this[storeKey].forEach(propKeyOrFunction => {
+    ;[...(this[protoStoreKey] || []), ...(this[instStoreKey] || [])].forEach(propKeyOrFunction => {
         const prop =
             typeof propKeyOrFunction === "string" ? this[propKeyOrFunction] : propKeyOrFunction
         if (prop !== undefined && prop !== null) {
@@ -20,7 +17,6 @@ function runDisposersOnWillUnmount() {
             prop()
         }
     })
-    this[storeKey] = []
 }
 
 export function disposeOnUnmount(target, propertyKeyOrFunction) {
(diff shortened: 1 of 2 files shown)


src/disposeOnUnmount.js
```diff
@@ -38,9 +34,16 @@ export function disposeOnUnmount(target, propertyKeyOrFunction) {
         )
     }
 
+    // decorator's target is the prototype, so it doesn't have any instance properties like props
+    const isDecorator = !target.hasOwnProperty("props")
```
mayorovp: Very fragile check. What if in future react devs will decide to move props to prototype?

JabX: I agree that it isn't the best check ever but I don't know what else we can do to discriminate usage. Also, `props` is pretty much the one property that will always be tied to the instance (`props` on the prototype would mean that all instance would share the same props, which completely defeats their purpose), so I think that, even in the future, we're fine making that assumption. Moreover, since classes are pretty much "legacy" now, I doubt anything will change for them in the future, and even if this for whatever reason change, we could still change this check for something else at that (very improbable) moment.

Anyway, now that v6 is out, this PR is pretty much obsolete for me (I had hoped that it could have been merged before) since the new restrictions on the method/decorator are breaking a good chunk of my uses for it. I'll still try to update it when I have the time though

FredyC:
@JabX do you want to close this PR if it's not relevant anymore?

src/disposeOnUnmount.js
mweststrate: Merging this one now, since it is relevant again after 6.10 :)

Replaced this specificy check with `typeof propertyKeyOrFunction === "string"`, since only when used as decorator we get a string here.

mweststrate:
Releasing as 6.1.1

src/disposeOnUnmount.js
JabX: This was my first idea too (and the one I thought about when I wrote the issue), but nothing prevents you to use `disposeOnUnmount(this, "myProperty")`... though I agree this isn't the way it's supposed to be used.

Anyway thanks a lot for merging the PR!

mweststrate:
Fair point! Will keep it in mind if it ever comes back to haunt us :)

PR #682, opened 2019-05-30
Fix Object.keys with ES6 syntax as props

### LINKED ISSUE #678: bug: ES6 syntax "static variable async vs static variable = async"
Thanks for the update and maintaining the project.

In the last v6.0.0
Applying @inject, static properties are not copied correctly if its defined in this syntax
```

@inject('store')
@observer
class Btn extends Component {
  // ❌ not working
  static async getInitialProps () {
    console.log("static");
  };
  render() {}
}
```
When you change it to variable based it works
```

@inject('store')
@observer
class Btn extends Component {
  // ✅this works
  static getInitialProps = async () => {
    console.log("static");
  };
  render() {}
}
```
My guess is the issue is in ```copyStaticProperties``` 

Could you guys have a look at this as it breaks any components now.
I am using it in nextjs as reference here.

Thank you and keep up the awesome work 👍

meabed:
- Changed .keys to .getOwnPropertyNames - This gets inherited or prototype properties such as "length, prototype,..."
- Filter properties that are inherited prototypes .getPrototypeOf
- Then copy the needed keys to target

- Fixes #678 
- Example on codesandbox here https://codesandbox.io/embed/kind-gauss-pfv5q

### DIFF
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -66,9 +66,14 @@ const hoistBlackList = {
 }
 
 export function copyStaticProperties(base, target) {
-    Object.keys(base).forEach(key => {
-        if (base.hasOwnProperty(key) && !hoistBlackList[key]) {
-            Object.defineProperty(target, key, Object.getOwnPropertyDescriptor(base, key))
-        }
-    })
+    const protoProps = Object.getOwnPropertyNames(Object.getPrototypeOf(base));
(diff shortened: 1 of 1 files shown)


mweststrate:
Thanks!

mweststrate:
Released as 6.0.3

PR #694, opened 2019-06-09
Add warning about cDU not firing

### LINKED ISSUE #692: [Bug?] componentDidUpdate() doesn't trigger after version 6
Like title says, componentDidUpdate() won't trigger since mobx-react version 6.

For example, I want to keep the scroll position to the bottom:

```
@inject('someStore')
@observer
class SomeList extends React.PureComponent {
...
componentDidMount() {
  this.scrollToBottom();
}

componentDidUpdate() { // never triggers, even though it re-renders, when there's new data
  this.scrollToBottom(); 
}
....
```

Trying with reaction, doesn't really solve the issue, since it doesn't wait for the element to render and it will scroll to the second to last element instead.

```
componentDidMount() {
  reaction(
    () => someStore.someArray.length, // or prop, etc..
   (val, currentReaction) => {
     this.scrollToBottom();
     this.scrollbarReactionToDispose = currentReaction; 
     // not sure if I need to dispose reaction in componentWillUnmount
     // since it's in the same class and if it's the correct way to do so
   }
  );
}

componentWillUnmount() {
  this.scrollbarReactionDispose.dispose();
}
```

I've also tried going back to a React.Component, just to be sure, but no luck.
I don't know if I'm missing something.

Thanks.

FredyC:
Closes #692


mweststrate:
Closed because of #703

PR #695, opened 2019-06-10
[V5] Warn about cDU not firing

FredyC:
Similar to #694 (issue #692), but backported to V5 code so people can actually prepare for an upgrade to V6.

PS: Made `v5` branch out of the latest V5 release so we can backport fixes that comes in V6.


mweststrate:
Thanks! But I think we can close this now due to #703

Issue #702, opened 2019-06-13
disposeOnUnmount not working on prototype


jer-sen:
When I use `disposeOnUnmount ` as a decorator on a property of a `React.Component` with React Native, I get the error `Attempting to change configurable attribute of unconfigurable property.`

A workaround is to change this line https://github.com/mobxjs/mobx-react/blob/00a3f94d52f929472c30be6fa6c42fb469d81db6/src/disposeOnUnmount.js#L86 with `configurable: true`.

`
react@16.8.3
react-native@0.59.0
mobx-react-devtools@6.1.1
`

mweststrate:
This might be fixed once #703 is merged

mweststrate:
Please check if 6.1.0 solves the problem.

FredyC:
Closing as it should be fixed. If the problem persist, please open a new issue with reproduction on the latest package version.

Issue #697, opened 2019-06-11
Adding `observer` affects original component


sergei-startsev:
**What is the current behavior?**
Adding `observer` wrapper affects original `render` method. There's some regression after upgrading `mobx-react`  to version **6** caused by affecting the original (**unwrapped**) component.

```jsx
class Foo extends React.Component {
  render() {
    //...
  }
}

const observableFoo = observer(Foo);
```
`observableFoo.render` as well as original `Foo.render` is overridden by adding  
```jsx
<Observer>
    <Component />
</Observer>
```
Since `observer` isn't pure, there's no possibility to reuse a component without mobx.

Here's a repository with repro steps: https://github.com/sergei-startsev/mobx-react-observer

**What is the expected behavior?**

There's an expectation that adding `observer` wrapper should not affect the original component. E.g.  adding `observer` wrapper in the default export below should not affect named `Foo` export: 
```jsx
export class Foo extends React.Component {
  // ...
  render() {
  }
}
export default inject('store')(observer(Foo));


// Foo shouldn't be wrapped
import { Foo } from './Foo.js';
```

FredyC:
I don't follow what's the issue. If you use exported `Foo`, it's unaffected by observer and as such can be reused without knowing anything about MobX. If you use default export in tests, it will have Observer in there, otherwise, it won't be there. I wouldn't consider it impure. Using the same input gives the same output.

Can you elaborate more please?

urugator:
Seems to be done deliberately, check this PR
https://github.com/mobxjs/mobx-react/commit/d3820da16e45bf2b97695bb465605cb2d2faea23#diff-3035dd5a37335cc50db60f91ad838797
Maybe there is some reasoning in the original PR:
https://github.com/mobxjs/mobx-react/pull/523
And it's also mentioned in this comment (last sentence):
https://github.com/mobxjs/mobx-react/issues/588#issuecomment-436228990

sergei-startsev:
@FredyC 
>If you use exported Foo, it's unaffected by observer and as such can be reused without knowing anything about MobX.

It is affected, take a look at `Foo > shallow` test from [repro steps](https://github.com/sergei-startsev/mobx-react-observer). It uses named unwrapped `Foo` export, however its snapshoot contains Observer component:
```diff
- <div>
-   Test
- </div>
+ <Observer>
+   <Component />
+ </Observer>
```

urugator:
Is this a viable workaround?
[...]

Issue #645, opened 2019-02-11
[On hold] Support concurrent mode React


mweststrate:
Support concurrent mode React. Primary open issue, don't leak reactions for component instances that are never committed.

A solution for that is described in: https://github.com/mobxjs/mobx-react-lite/issues/53

But let's first wait until concurrent mode is a bit further; as there are no docs yet etc, introducing a workaround in the code base seems to be premature at this point (11-feb-2019)

mweststrate:
That is: assuming no one is using concurrent mode in production atm

FredyC:
For anyone interested `mobx-react-lite@next` has experimental support for Concurrent mode. Feel free to test it out.

Personally, I don't think it's worth the hassle supporting Concurrent mode in class components.

bsmith-cycorp:
They just published a post getting further into details about Concurrent Mode: https://reactjs.org/docs/concurrent-mode-intro.html

FredyC:
I believe we are ready for Concurrent mode with `mobx-react-lite@next`. Of course, at some point, we shall try against "experimental" version just to be sure. PR welcome as well :)

bsmith-cycorp:
I hadn't heard about `mobx-react-lite`.

Is it intended to supercede `mobx-react`? That would be hugely disappointing for me, as I take serious philosophical issue with hooks, especially given MobX's emphasis on (and synergy with) classes.

FredyC:
Um no, `mobx-react@6` is built on top of `mobx-react-lite` :)

mweststrate:
Ok, FWIW, using suspense with observables is pretty natural if you `throw when(() => expr)` :)

https://codesandbox.io/s/dry-microservice-qnle3

FredyC:
@mweststrate That's a certainly nice feat, although I never liked much to have fetching logic inside the stores. But I guess with Suspense it will make more sense to have it separated from components for a proper preloading.

I also love that it will re-render on further updates to those observables without any extra magic. That makes it really powerful.

That `fetchingPosts` extra state is awkward though, considering Suspense should kinda remove the worry about any sort of loading state, it shouldn't be needed. I do understand why it's there in this case, but it's not ideal for sure.

mweststrate:
Suspense doesn't make any decisions on who triggers rendering. So I didn't
do either in the example.

THe first is triggered 'externally', the second is triggered by rendering,
just to demonstrate both approaches.

The loading state again is a design decision. Should every render trigger a
[...]

PR #703, opened 2019-06-17
Restored mixin based implementation of `observer` for class based components

### LINKED ISSUE #699: mobx-react 6 shallow testing regression
**What is the current behavior?**

 With `mobx-react` 6 upgrade regressions are detected in unit tests:

```jsx
// Foo.jsx
class Foo extends React.Component {
  //...
  render() {
    return <div>Test</div>;
  }
}
export default inject('store')(observer(Foo));
```
```jsx
//Foo.test.jsx
describe('Foo', () => {
  test('shallow', () => {
    const component = shallow(<Foo.wrappedComponent />);
    expect(component).toMatchSnapshot();
  });
})
```
Snapshot:
```jsx
<Observer>
  <Component />
</Observer>
```
Here's the repository with repro steps: https://github.com/sergei-startsev/mobx-react-6-regression

**What is the expected behavior?**

Expected snapshot:
```jsx
<div>
  Test
</div>
```
---
I believe the issue is caused by added `<Observer>` wrapper:
https://github.com/mobxjs/mobx-react/blob/4904b624f46be2e14d3a3d1017c1fa9bcb334f02/src/observer.js#L108

See https://github.com/mobxjs/mobx-react/commit/1ea40e21117d9c30067dd7b0267aa80b99473c59#diff-08f3e9ab539253c87ea111cc61c69e29R156.

mweststrate:
Proposal to revert to the classic implementation for class based observer components; by monkey patching the prototype rather than wrapping render in `<Observer>`

Fixes / affects #699, #692, #697, #702

### DIFF
--- a/src/observerClass.js
+++ b/src/observerClass.js
@@ -1,9 +1,6 @@
 import { PureComponent, Component } from "react"
 import { createAtom, _allowStateChanges, Reaction, $mobx } from "mobx"
-import {
-    useStaticRendering as useStaticRenderingLite,
-    isUsingStaticRendering
-} from "mobx-react-lite"
+import { isUsingStaticRendering } from "mobx-react-lite"
 
 import { newSymbol, shallowEqual, setHiddenProp, patch } from "./utils/utils"
 
@@ -11,10 +8,6 @@ const mobxAdminProperty = $mobx || "$mobx"
 const mobxIsUnmounted = newSymbol("isUnmounted")
 const skipRenderKey = newSymbol("skipRender")
 const isForcingUpdateKey = newSymbol("isForcingUpdate")
-
-export function useStaticRendering(useStaticRendering) {
-    useStaticRenderingLite(useStaticRendering)
-}
 
 export function makeClassComponentObserver(componentClass) {
     const target = componentClass.prototype
@@ -51,29 +44,6 @@ export function makeClassComponentObserver(componentClass) {
 function makeComponentReactive(render) {
     if (isUsingStaticRendering() === true) return render.call(this)
 
-    function reactiveRender() {
-        isRenderingPending = false
-        let exception = undefined
-        let rendering = undefined
-        reaction.track(() => {
-            try {
-                rendering = _allowStateChanges(false, baseRender)
-            } catch (e) {
-                exception = e
-            }
-        })
-        if (exception) {
-            throw exception
-        }
-        return rendering
-    }
-
-    // Generate friendly name for debugging
-    const initialName =
-        this.displayName ||
-        this.name ||
-        (this.constructor && (this.constructor.displayName || this.constructor.name)) ||
-        "<component>"
     /**
      * If props are shallowly modified, react will render anyway,
      * so atom.reportChanged() should not result in yet another re-render
@@ -85,8 +55,14 @@ function makeComponentReactive(render) {
      */
     setHiddenProp(this, isForcingUpdateKey, false)
 
-    // wire up reactive render
+    // Generate friendly name for debugging
+    const initialName =
+        this.displayName ||
+        this.name ||
+        (this.constructor && (this.constructor.displayName || this.constructor.name)) ||
+        "<component>"
     const baseRender = render.bind(this)
+
     let isRenderingPending = false
 
     const reaction = new Reaction(`${initialName}.render()`, () => {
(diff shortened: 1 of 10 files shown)


src/observerClass.js
```diff
@@ -0,0 +1,159 @@
+import { PureComponent, Component } from "react"
+import { createAtom, _allowStateChanges, Reaction, $mobx } from "mobx"
+import { useStaticRendering as useStaticRenderingLite } from "mobx-react-lite"
+
+import { newSymbol, shallowEqual, setHiddenProp, patch } from "./utils/utils"
+
+let isUsingStaticRendering = false
```
FredyC: Do you actually need to keep separate state for this? Why not reuse the variable from the Lite directly and export that setter directly as well?

mweststrate: Because an imported variable is not bound to its original definition (that is, changes won't be reflected in it)

FredyC: I don't think it should matter. The Lite exports both setter and getter. It does not export the actual variable. For consistency, the same should be exported here as well imo.

mweststrate: ah missed that, will change!

mweststrate: Done

urugator:
Huh, haven't noticed that patching logic became so convoluted over time.
I know it's been mentioned many times that patching is necessary, but thb I am not sure about specifics. 
What's exactly problematic about inheriting the original component? Like:      
```javascript
class extends originalComponent {
  componentWillUnmount() {
    super();
    // dispose 
  }
}
```

mweststrate:
Yeah, it's all disposeOnUnmount's fault :-(. It introduces a plethorea of
possiblities: the decorator want to set CWU, so does observer, so does the
component as defined by the user (either on prototype or by using an arrow
function). So the thing is monkey patches on different levels, which is
especially annoying if multiple inheritance trees are involved. See also
#705 that tries to simplifies things again, but breaks some cases

src/observerClass.js
FredyC: Nit: The observerClass should only consume `isUsingStaticRendering`, but both should be exported directly from `index.js` without extra wrapping.

src/observerClass.js
```diff
@@ -0,0 +1,161 @@
...
+     */
+    setHiddenProp(this, skipRenderKey, false)
+    /**
+     * forceUpdate will re-assign this.props. We don't want that to cause a loop,
+     * so detect these changes
+     */
+    setHiddenProp(this, isForcingUpdateKey, false)
+
+    // wire up reactive render
+    const baseRender = render.bind(this)
```
FredyC: Perhaps move this line above `reactiveRender` declaration? It's fairly confusing to follow this code with all that hoisting. I would even move `const reaction` on top for a natural flow.

src/observerClass.js
mweststrate: Fixed

src/observerClass.js
mweststrate: yeah it is only the `disposeOnUnmount` decorator support that introduces the actual complexity, not the class support itself (which is pretty simple, and doesn't need the ugly component-might-not-have-been-committed hack :))

urugator: > doesn't need the ugly component-might-not-have-been-committed hack

Are you sure? I think the reaction would have to be created in `componentDidMount` (similary to inside `useEffect`).
Check https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html#adding-event-listeners-or-subscriptions:
Specifically:
> async rendering (where rendering might be interrupted before it completes, causing componentWillUnmount not to be called).

mweststrate:
Dang, that is a good point. Would you mind opening an issue for that? IIRC
class components don't support this rendered-once but never comimtted
behavior. But in hindsight I'm not sure that actually makes sense.

urugator:
Well technically there is a one already #645 :D

mweststrate:
Fair point!

urugator:
And I have another bad news ... this also affects `disposeOnUnmount`... actually `@disposeOnUnmount` is conceptually completely wrong, because constructors shouldn't contain side effects... all side effects should be in `cDM` or `cDU`.

mweststrate:
Yeah I think we have to work to a hard split in the future: If you want to
use concurrent mode, limit yourself mobx-react-lite. If that is too hard of
a split, we could make v7 a minimal class implementation (as proposed
already) without all the other shizzle: Provider / inject / propTypes /
disposeOnUnmount

PR #705, opened 2019-06-18
Test cleanup

mweststrate:
Depends on #703 

This PR removes the usage of create-react-class


mweststrate:
N.B. diff is distorted until #703 is merged

Issue #489, opened 2018-06-01
Adding @observer to a class that extends a class already wrapped in @observer


skiritsis:
When adding `@observer` to a class that extends a class already wrapped in `@observer` you get the following warning:
`Use `shouldComponentUpdate` in an `observer` based component breaks the behavior of `observer` and might lead to unexpected results. Manually implementing sCU should not be needed when using mobx-react. `

I understand that this change was made in version 5.1.0 as mentioned in changelog:
> Mobx-react will now print a warning when combining observer with a custom shouldComponentUpdate implementation. Fixes #417 

My question is, if this is a valid scenario that users should be warned about and can potentially lead to problems or is the warning is a false positive? The 2nd component detects the `sCU` implemented by the parent component and displays the warning(which will become an exception in future versions).

In my case both components implement the render method, and both of them are used in different parts of the application so both of them need the `@observer` wrapper.

Adding also a codepen [example](https://codesandbox.io/s/p59mj5n840) displaying the problem.

urugator:
Use of inheritance is discouraged, if inavoidable for whatever reason, decorate only "leaf" classes. 
So in your case introduce non-decorated `AbstractHello` (making clear it shouldn't be instantiated on it's own) and decorated  `Hello1`, `Hello2` both extending from `AbstractHello`.
However I think you may still encounter all sort of issues, especially if parent class implements lifecycle methods. Consider using HoC, render callbacks etc instead.

JabX:
I have the same issue and I already gave some thought about how to refactor the classes without inheritance, but in my case I'm pretty sure that it's the easiest way to proceed (derived class is a specialized version of the base class that overrides some of its behaviors and uses a lot of its state).

Anyway, looking at [this](https://github.com/mobxjs/mobx-react/blob/master/src/observer.js#L225) and how simple the "reactiveMixin" is, I don't think that there is any problem with subclassing observer components. The warning in question is even totally pointless because the SCU that's defined on the component is already the exact same function we're trying so set. So a very simple fix could be to add a `target.shouldComponentUpdate === reactiveMixin.shouldComponentUpdate` test before, so that we can continue subclassing our components in peace. Unless I'm missing something, of course.

mweststrate:
Sounds like a sane fix indeed, PR with test and proposed change is welcome
:).

Op vr 1 jun. 2018 om 16:04 schreef Damien Frikha <notifications@github.com>:

mweststrate:
Closing as MR above has been released

PR #707, opened 2019-06-18
[WIP] cleaner disposeOnCleanup implementation

mweststrate:
Depends on #704, #705 

This PR reimplements disposeOnCleanup. This is a breaking change. Goal: simplify the implementation of dispose on cleanup (as we tried original in v6, before the old implementation of observer was reverted in #703)

This PR changes the rules of the game of disposeOnCleanup a bit:
- Can only be used on `observer` based components
- `observer` cannot be applied multiple times in an inheritance tree (avoids certain edge cases)
- `render` and `componentWillUnmount` will throw if written to (avoids problems where people use an arrow function with field initializer, instead of declaring the method properly on the prototype)
- running disposers is now done as part of the cleanup of `observer`
- This should all simplify the implementation significantly. However, since some constraints are changed, this is a breaking change

Todo:
- [ ] fix / update all tests
- [x] restore applicable tests from v5
- [x] check tests and implementation #671, things might be applicable here from that PR

### DIFF
--- a/src/disposeOnUnmount.js
+++ b/src/disposeOnUnmount.js
@@ -1,16 +1,33 @@
 import * as React from "react"
-import { patch, newSymbol } from "./utils/utils"
+import { newSymbol } from "./utils/utils"
 
 const protoStoreKey = newSymbol("disposeOnUnmountProto")
 const instStoreKey = newSymbol("disposeOnUnmountInst")
 
-function runDisposersOnWillUnmount() {
+function checkFunc(prop) {
+    if (typeof prop !== "function") {
+        throw new Error(
+            "[mobx-react] disposeOnUnmount only works on functions such as disposers returned by reactions, autorun, etc."
+        )
+    }
+}
+
+function check(prop) {
+    if (Array.isArray(prop)) {
+        prop.map(checkFunc)
+    } else {
+        checkFunc(prop)
+    }
+}
+
+export function runDisposersOnWillUnmount() {
     ;[...(this[protoStoreKey] || []), ...(this[instStoreKey] || [])].forEach(propKeyOrFunction => {
         const prop =
             typeof propKeyOrFunction === "string" ? this[propKeyOrFunction] : propKeyOrFunction
         if (prop !== undefined && prop !== null) {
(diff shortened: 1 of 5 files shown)


JabX:
> `observer` cannot be applied multiple times in an inheritance tree (avoids certain edge cases)

I am not very sure about how `observer` works with inheritance today (at least in v5), but I know putting `observer` everywhere works so this is what I do. I have a main `observer` class component, as well as an inherited component that is overrides a few methods. Since I can't put `observer` on both anymore, does everything work if only the parent class gets the `observer` decorator? I am mostly interested in observable props and an overridden render method. I guess the worst case scenario would be to separate the observer'd/non-observer'd parent class and inherit from the latter, which could be manageable but not ideal I guess.

mweststrate:
It (should) mean that observer on either base or subclass should suffice.

JabX:
In that case, is it possible to contribute some tests to confirm that behavior? We already had some unwanted breaking changes in the past about this (see #489), so I'd like some confirmation that it won't break again (without notice) in the future. I understand that it is an edge case and not exactly recommended in the general case (so it probably doesn't belong in the docs), but at least some tests could show what works and what doesn't.

mweststrate:
Yes, always, and awesome! Finding out how things are *actually* being used
is always the hardest part.

t0nyh0:
Is there an ETA for when this will be merged?  We're stuck on 5.0.0 until this PR lands.  We've tried using the latest version (6.1.3) but it appears that the patched `componentWillUnmount`" function is not executing so we're ending up with a lot of "forceUpdate warnings" and the render reaction being disposed.

mweststrate:
If componentWIllMount is not executed, that is most probably unrelated to
this.

Either:
1. Don't use Enzyme's shallow rendering
2. Or: make sure you don't declare render / lifecycle hooks as arrow
functions.

If that doesn't work, please file a fresh issue instead, I'm not convinced
yet these things are related

mweststrate:
This PR is a year old and it seems nobody really missed it. Do it manually or use hooks instead :-P. Closing.

PR #711, opened 2019-06-19
Add the displayName static property to the Provider component

vkrol:
The `Provider` component has lost its descriptive name in React Dev Tools in the `6.x` version of `mobx-react`:
Снимок экрана 2019-06-19 в 23 32 13
This PR adds it back:
Снимок экрана 2019-06-19 в 23 33 55

### DIFF
--- a/src/Provider.js
+++ b/src/Provider.js
@@ -7,6 +7,7 @@ export const MobXProviderContext = createContext({})
 
 export class Provider extends Component {
     static contextType = MobXProviderContext
+    static displayName = "MobXProvider"
 
     constructor(props, context) {
         super(props, context)


src/Provider.js
```diff
@@ -7,6 +7,7 @@ export const MobXProviderContext = createContext({})
 
 export class Provider extends Component {
     static contextType = MobXProviderContext
+    static displayName = "Provider"
```
FredyC: Perhaps `MobXProvider` might be a more useful name to distinguish from other providers?

vkrol: I thought about it, but decided to stay at the more conservative decision. If you think that this needs to be changed, then I will do it.

FredyC: Please do.

vkrol: Done.

vkrol:
@FredyC CI checks were not passed :(

PR #715, opened 2019-06-20
Add the displayName static property to the Injector component

vkrol:
The `Injector` component has lost its descriptive name in React Dev Tools in the 6.x version of mobx-react:
Снимок экрана 2019-06-20 в 20 55 56
This PR adds it back:
Снимок экрана 2019-06-20 в 20 57 30

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -11,6 +11,7 @@ function createStoreInjector(grabStoresFn, component, injectNames, makeReactive)
 
     class Injector extends Component {
         static contextType = MobXProviderContext
+        static displayName = displayName
 
         render() {
             const { forwardRef, ...props } = this.props


FredyC:
Well, that doesn't seem correct. There is already a `inject(App)` present once. The second one should be `App` only, shouldn't it?

vkrol:
@FredyC No, `App` is in the second `inject(App)` (it is collapsed on the screenshot). Second `inject(App)` was added in `mobx@6.x` because of the added support of `forwardRef`.

FredyC:
I don't know, feels perhaps slightly more confusing to have a component with the exact same name twice there. I don't think it will do any harm to keep minified name there since everything important is already visible.

I will leave it open for others to chime in.

vkrol:
I did it under the impression of how it is done in the latest version of `react-redux`. I think that minified name is more confusing than this. I am ready to consider other more understandable options.

vkrol:
I think that the minified name is bad because it is easy to confuse it with some application component, not library component.

vkrol:
@FredyC I can suggest the following option:
```
<MobXProvider>
  <Context.Provider>
    <ForwardRef(inject(App))>    <-----
      <inject(App)>
         <App>
           ...
```

FredyC:
Hm, looking at that code I actually wonder if that class component needs to be there. Converting a whole thing to a single functional component would yield tree-like.

```
<MobXProvider>
  <Context.Provider>
      <inject(App)>
         <App>
```

vkrol:
@FredyC how will this functional component consume the React context?
https://github.com/mobxjs/mobx-react/blob/master/src/inject.js#L13

FredyC:
Um, with `useContext`? :) Since V5 depends on React 16.8, we can use Hooks internally just fine.

vkrol:
> Since V6 depends on React 16.8, we can use Hooks internally just fine.
Thanks! I am forgot about this.

FredyC:
@vkrol Feeling up for the PR? I might do it in a couple of days, but if you want some hooks training, go ahead :)

vkrol:
@FredyC thanks, I will try it today or tomorrow 😀

vkrol:
> Hm, looking at that code I actually wonder if that class component needs to be there. Converting a whole thing to a single functional component would yield tree-like.

Done.
Fixed by https://github.com/mobxjs/mobx-react/pull/716.

FredyC:
6.1.2 published with this

PR #716, opened 2019-06-22
Get rid of the redundant Injector wrapper

### LINKED ISSUE #717: Inject does not work correctly with components that use React.forwardRef
`inject` does not work correctly with components that use `React.forwardRef`. 
`isStateless` returns `true` for components that wrapped via `React.forwardRef` (https://github.com/mobxjs/mobx-react/blob/master/src/inject.js#L20) so `this.props.forwardRef` is not passed to these components.
```
const React = require("react");
const ReactDOM = require("react-dom");
const { inject, Provider } = require("mobx-react");

const e = React.createElement;

test("inject should work with components that use forwardRef", () => {
    const FancyComp = React.forwardRef((_, ref) => {
        return e("div", { ref });
    });
    
    const InjectedFancyComp = inject("bla")(FancyComp);
    const ref = React.createRef();
    
    ReactDOM.render(
        e(
            Provider,
            { bla: 42 },
            e(
                InjectedFancyComp,
                { ref },
                )
            ),
        document.createElement("div")
    );
    
    expect(ref.current).not.toBeNull(); // this assertion is failing
    expect(ref.current).toBeInstanceOf(HTMLDivElement);
});
```
Minimal reproducible example: https://github.com/vkrol/mobx-react-inject-forward-ref-bug

vkrol:
This is the proof of concept:
> Hm, looking at that code I actually wonder if that class component needs to be there. Converting a whole thing to a single functional component would yield tree-like.

https://github.com/mobxjs/mobx-react/pull/715#issuecomment-504358745

Tests are green, but I did not check it on a real production codebase.

Fixes #717 too.

### DIFF
--- a/src/inject.js
+++ b/src/inject.js
@@ -1,41 +1,33 @@
 import React, { Component, createElement } from "react"
 import { observer } from "./observer"
-import { isStateless, copyStaticProperties } from "./utils/utils"
+import { copyStaticProperties } from "./utils/utils"
 import { MobXProviderContext } from "./Provider"
 
 /**
  * Store Injection
  */
 function createStoreInjector(grabStoresFn, component, injectNames, makeReactive) {
-    let displayName = getInjectName(component, injectNames)
+    // Support forward refs
+    let Injector = React.forwardRef((props, ref) => {
+        const newProps = { ...props }
+        const context = React.useContext(MobXProviderContext)
+        Object.assign(newProps, grabStoresFn(context || {}, newProps) || {})
 
-    class Injector extends Component {
-        static contextType = MobXProviderContext
-
-        render() {
-            const { forwardRef, ...props } = this.props
-
-            Object.assign(props, grabStoresFn(this.context || {}, props) || {})
-
-            if (forwardRef && !isStateless(component)) {
-                props.ref = this.props.forwardRef
-            }
-
-            return createElement(component, props)
+        if (ref) {
+            newProps.ref = ref
         }
-    }
+
+        return createElement(component, newProps)
+    })
+
     if (makeReactive) Injector = observer(Injector)
     Injector.isMobxInjector = true // assigned late to suppress observer warning
 
-    // Support forward refs
-    const InjectHocRef = React.forwardRef((props, ref) =>
-        React.createElement(Injector, { ...props, forwardRef: ref })
-    )
     // Static fields from component should be visible on the generated Injector
-    copyStaticProperties(component, InjectHocRef)
-    InjectHocRef.wrappedComponent = component
-    InjectHocRef.displayName = displayName
-    return InjectHocRef
+    copyStaticProperties(component, Injector)
+    Injector.wrappedComponent = component
+    Injector.displayName = getInjectName(component, injectNames)
+    return Injector
 }
 
 function getInjectName(component, injectNames) {
--- a/src/utils/utils.js
+++ b/src/utils/utils.js
@@ -1,9 +1,3 @@
-export function isStateless(component) {
-    // `function() {}` has prototype, but `() => {}` doesn't
-    // `() => {}` via Babel has prototype too.
-    return !(component.prototype && component.prototype.render)
-}
-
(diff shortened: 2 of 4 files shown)


src/inject.js
```diff
@@ -7,35 +7,27 @@ import { MobXProviderContext } from "./Provider"
...
-    class Injector extends Component {
-        static contextType = MobXProviderContext
-
-        render() {
-            const { forwardRef, ...props } = this.props
+    // Support forward refs
+    let Injector = React.forwardRef((props, ref) => {
+        const newProps = { ...props }
+        const context = React.useContext(MobXProviderContext)
+        Object.assign(newProps, grabStoresFn(context || {}, newProps) || {})
```
FredyC: Slightly more concise and done within a single operation. You can drop a line 12.
```suggestion
        const nextProps = Object.assign({}, props, grabStoresFn(context || {}, props) || {})
```

src/inject.js
```diff
@@ -7,35 +7,27 @@ import { MobXProviderContext } from "./Provider"
...
 
-            Object.assign(props, grabStoresFn(this.context || {}, props) || {})
+        if (ref && !isStateless(component)) {
```
FredyC: What about moving this condition outside and basically avoid calling `forwardRef` for a non-stateless component?

Although I do wonder what's the purpose of that. Why is the ref passed to class components only? What am I missing here?

src/inject.js
vkrol: We cannot do it because `grabStoresByName` modifies `props` passed as the second argument https://github.com/mobxjs/mobx-react/blob/master/src/inject.js#L66.

src/inject.js
vkrol: > What about moving this condition outside and basically avoid calling forwardRef for a non-stateless component?

I will try it.

> Although I do wonder what's the purpose of that. Why is the ref passed to class components only? What am I missing here?

Because function components cannot be given refs?

src/inject.js
FredyC: I see. Ok, let's leave as you did it. Not worth an extra hassle.

src/inject.js
FredyC: > Because function components cannot be given refs?

True.

Btw, might as well rename `isStateless` to `isFunctional` to stay current :)

vkrol: > What about moving this condition outside and basically avoid calling forwardRef for a non-stateless component?

I’m having trouble trying to figure out how to do it in a simple way without repetition.
https://gist.github.com/vkrol/e2804b367c28f96b0802f21fa1845c18

vkrol:
I deleted `isStateless` and now we apply `React.forwardRef` unconditionally.

> Can you also add the test case you drafted here in that PR?

I will do it tomorrow because I have some problems with `mobx-react` tests.

vkrol:
> I will do it tomorrow because I have some problems with mobx-react tests.

Done.

src/inject.js
```diff
@@ -1,41 +1,33 @@
...
-        render() {
-            const { forwardRef, ...props } = this.props
+    // Support forward refs
+    let Injector = React.forwardRef((props, ref) => {
+        const newProps = { ...props }
+        const context = React.useContext(MobXProviderContext)
+        Object.assign(newProps, grabStoresFn(context || {}, newProps) || {})
 
-            Object.assign(props, grabStoresFn(this.context || {}, props) || {})
+        if (ref) {
```
FredyC: I wonder if that condition is actually needed. Worst case scenario the `undefined` gets assigned here and that's like nothing happened. That means you can just stick that ref to line 12, right?

vkrol: We cannot do it because it will be `null`, not `undefined` and it will trigger the following warning for function components:
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
```

FredyC:
Well, nobody else seems to care, so let's ship it and fix later if necessary :)

PR #720, opened 2019-06-26
Observer checks for use of React.memo

### LINKED ISSUE #709: 'componentWillReact' of undefined after updating to 6.1.0.
Due to the recent `componentDidUpdate` bug in mobx-react v6 we tried to upgrade to 6.1.0 but the page doesn't load and got this in the console.

```
vendors.bundle.a88a35b9.js:145824 Uncaught TypeError: Cannot read property 'componentWillReact' of undefined
```

"mobx": "^4.5.0",
"mobx-react": "^6.1.0"

Tried updating mobx to 4.11.0 and mobx-react to 6.1.1 doesn't fix it either.

FredyC:
Fixes #709 

The `React.memo` returns an object which isn't expected and is passed down to `observerClass` which assumed a class component, but it got a plain object without a prototype present.

The solution works in a similar fashion as the one for `React.forwardRef`, but it simply removes additional `React.memo` by accessing a wrapped function component and applying `observer` to it. And it issues a warning.

---

I've also installed `@testing-library/react` and `jest-mock-console` to start modernizing tests slightly. There was a small hiccup as we have TS 2.6 in here, but the latest peer dep `@types/yargs` is built on top of TS 3.5 (requiring `unknown`). I had to force the older version there.


FredyC:
> I would prefer it to throw.

The throw would probably make sense if it wasn't possible to recover from it.

urugator:
I don't want to argue about this, but imo either it's a valid use case then it shouldn't warn (at least on production) or it's not and then it should throw (so I am forced to fix it).

mweststrate:
Is it allowed to use static members on memo components? Such as `defaultProps`, `propTypes` etc? If that is the case, we should throw instead of fixing, as we might accidentally loose those members.

urugator:
It is https://github.com/facebook/react/pull/14298#issue-232722530

FredyC:
Ok, you are right. Changed it to throw. Sadly, the test is worthless in that case, but I kept it there for a future reference when it's possible to properly test errors in React.

FredyC:
I think it's time to merge :)

Issue #677, opened 2019-05-29
Shared user guide


FredyC:
@mweststrate With 6.0 out now (👏), would you consider trimming down README and consolidate that at the docs site I've made? I still haven't got any feedback about that site from you.

https://mobx-react.js.org

mweststrate:
Sorry, still have to look into it, but that will at earliest be next week.

Op wo 29 mei 2019 23:29 schreef Daniel K. <notifications@github.com>:

mweststrate:
Only glanced over it, but some feedback points:

1. I think having a third place for documentation for mobx-react is a bad idea. Two (mobx.js.org and the readme of this repository itself) is basically already bad enough. I think the documentation proposed here should be a mobx+react recipes page on the official docs (between section 8.0 and 8.1?). And then forward to it from the `observer` page. Some of the sections have quite some overlap with the general documentation and could be skipped I think
2. Beware of being dismissive about class based components. We are not a React style police, and people will be stuck with them for the coming years in large existing code bases. There is no need to alienate those people just because they aren't running a greenfield project. The page / docs in general are written primarily from a mobx-react-lite perspective. That is ok, but the domain on which it is hosted suggests otherwise :) 	
3. I think the primary point should be to explain how to achieve certain goals that might developers have in certain situtations: such as external state, component local state, and side effects both with hooks and classes. And furthermore elaborate as done on the subtile differences between the HoC, hook and wrapper. 

Some of the example are a bit convoluted, for example on the introduction page:

Creating an inline component is quite a bad idea, as it kills the reconciler, and at least a bad example to set: `const renderTodo = (done: boolean) => todo => (`. I suggest to extract it to  a separate component and make it an observer. Which is consistent with the rest of the MobX docs: renderers for list items should be separate components

The localStore's addTodo is expressed in terms of the ref. If instead the text to be added would be passed in, there would be a clear separation between the logic and the internal stuff of the component, and the entire initializer could actually be extracted out of the component.

The example uses `createTodos`, but it seemst to be defined nowhere?

So I think there could be a bit simpler example to introduce the whole concept :). Random idea: https://codesandbox.io/s/evwj5. Yes, it is more contrived, but it will have less noise so it is easier to see what it is all about

---

All IMHO of course, awesome that you are taking the effort!

The beta notice on the libraries page can be removed. 

Sections I really liked:
* https://mobx-react.js.org/libraries
* https://mobx-react.js.org/state-outsourcing
[...]

PR #721, opened 2019-07-01
Add links to documentation site

FredyC:
A couple of links in readme leading to the site I've made. This is related to #677. It's certainly too soon to call it an official site for everyone (classes are not explained much in there), but a couple of links so people can start learning won't hurt if they want.

I am a bit doubtful about documenting the `useLocalStore` and `useAsObservableSource` here in such detail. Since these are pure hooks, they are well covered on that site already. Besides, it's already bit outdated here as it doesn't cover a second argument to `useLocalStore`. What do you think? Shall I remove it?


FredyC:
Alright, I guess not really a big change needing a review. Rolling it out :)

PR #724, opened 2019-07-10
Use react-testing-library Part 1

vkrol:
From https://github.com/mobxjs/mobx-react/pull/720#issue-292091888:
> I've also installed @testing-library/react and jest-mock-console to start modernizing tests slightly. 

This is the first part of conversion of the tests to `react-testing-library`.

### DIFF
--- a/test/inject.test.js
+++ b/test/inject.test.js
@@ -4,7 +4,7 @@ import * as mobx from "mobx"
 import { action, observable } from "mobx"
 import { observer, inject, Provider } from "../src"
 import { sleepHelper, withConsole } from "./index"
-import { render, cleanup } from "@testing-library/react"
+import { render } from "@testing-library/react"
 import renderer, { act } from "react-test-renderer"
 
 describe("inject based context", () => {
@@ -682,5 +682,3 @@ test("#612 - mixed context types", () => {
 
     render(<App />)
 })
-
-afterEach(cleanup)
--- a/test/inject.test.js
+++ b/test/inject.test.js
@@ -1,10 +1,10 @@
 import React, { Component } from "react"
 import * as PropTypes from "prop-types"
-import ReactDOM from "react-dom"
 import * as mobx from "mobx"
 import { action, observable } from "mobx"
 import { observer, inject, Provider } from "../src"
-import { createTestRoot, cleanupTestRoot, sleepHelper, withConsole } from "./index"
+import { sleepHelper, withConsole } from "./index"
+import { render, cleanup } from "@testing-library/react"
 import renderer, { act } from "react-test-renderer"
 
 describe("inject based context", () => {
@@ -394,7 +394,7 @@ describe("inject based context", () => {
         expect(ref2.current.didRender).toBe(true)
     })
 
-    test("inject should work with components that use forwardRef", done => {
+    test("inject should work with components that use forwardRef", () => {
         const FancyComp = React.forwardRef((_, ref) => {
             return <div ref={ref} />
         })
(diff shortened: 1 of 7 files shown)


test/stateless.test.js
```diff
@@ -58,8 +54,8 @@ test("stateless component with context support", async () => {
         </C.Provider>
     )
 
-    await asyncReactDOMRender(<ContextProvider />, testRoot)
-    expect(testRoot.querySelector("div").innerHTML.replace(/\n/, "")).toBe("context: hello world")
+    const { container } = render(<ContextProvider />)
+    expect(container.querySelector("div").innerHTML.replace(/\n/, "")).toBe("context: hello world")
```
FredyC: Shouldn't this work as well? There are more places like that in here. It's more readable imo.
```suggestion
    expect(getByText("context: hello world")).toBeTruthy()
```

test/stateless.test.js
```diff
@@ -38,8 +34,8 @@ describe("stateless component with propTypes", () => {
     })
 
     test("render test correct", async () => {
-        await asyncReactDOMRender(<StatelessCompObserver testProp="hello world" />, testRoot)
-        expect(testRoot.querySelector("div").innerHTML).toBe("result: hello world")
+        const { container } = render(<StatelessCompObserver testProp="hello world" />)
```
FredyC: Perhaps we can catch the drift and start calling these correctly a function components. Even rename a whole test file.

test/inject.test.js
```diff
@@ -1,10 +1,10 @@
 import React, { Component } from "react"
 import * as PropTypes from "prop-types"
-import ReactDOM from "react-dom"
 import * as mobx from "mobx"
 import { action, observable } from "mobx"
 import { observer, inject, Provider } from "../src"
-import { createTestRoot, cleanupTestRoot, sleepHelper, withConsole } from "./index"
+import { sleepHelper, withConsole } from "./index"
```
FredyC: Can we perhaps use `jest-mock-console` instead?

vkrol: It has nothing to do with `react-testing-library`. I can investigate it in the separate PR.

test/stateless.test.js
vkrol: It has nothing to do with `react-testing-library`. I can investigate it in the separate PR.

test/stateless.test.js
vkrol: Thanks, I will investigate this later today.

vkrol: I do not think that this is the good idea because getByText will throw an error if the element will not be found, so the assertion will be redundant. I simplified assertions a bit.

FredyC: Fair enough, doesn't matter in simple cases like that. However, in apps, I often do just `getByText` without expect because assertion is nothing else than throwing an error :) The output in case of failure does not differ that much.

vkrol: OK, thanks. I prefer assertions via `expect` because with them it becomes more clear what expectations we have in tests.

vkrol:
@FredyC All conversations have been resolved.

vkrol:
@FredyC Do I need to do something else?

FredyC:
Ah sorry, I thought you are still working on it :)

Just thinking I would move `afterEach(cleanup)` to the test/index.js instead of repeating in every test file. It's a generally a good assumption that each test should start with a clean slate.

Otherwise LGTM

vkrol:
@FredyC 
> Just thinking I would move afterEach(cleanup) to the test/index.js instead of repeating in every test file. 

Done.

FredyC:
Thanks. Looking forward to other parts :)

vkrol:
@FredyC Thanks! I'll do it in a few weeks.

PR #728, opened 2019-07-18
Use react-testing-library Part 2

vkrol:
This is the follow up to https://github.com/mobxjs/mobx-react/pull/724.


test/disposeOnUnmount.test.js
```diff
@@ -15,7 +13,7 @@ async function testComponent(C, afterMount, afterUnmount) {
         afterMount(cref)
     }
 
-    await asyncReactDOMRender(null, testRoot)
+    render(null, { container })
```
FredyC: What about using https://testing-library.com/docs/react-testing-library/api#unmount instead?

vkrol: Sure, thanks!

PR #731, opened 2019-07-28
Use react-testing-library Part 3

vkrol:
This is the follow up to https://github.com/mobxjs/mobx-react/pull/728.

### DIFF
--- a/test/misc.test.js
+++ b/test/misc.test.js
@@ -1,13 +1,11 @@
 import React, { Component } from "react"
-import ReactDOM from "react-dom"
 import * as mobx from "mobx"
 import { observer } from "../src"
-import { createTestRoot, withConsole } from "./index"
+import { withConsole } from "./index"
 import renderer, { act } from "react-test-renderer"
+import { render } from "@testing-library/react"
 
 const mobxAdminProperty = mobx.$mobx || "$mobx"
-
-const testRoot = createTestRoot()
 
 test("issue mobx 405", () => {
     function ExampleState() {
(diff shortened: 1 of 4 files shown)


test/misc.test.js
```diff
@@ -80,21 +78,14 @@ test("#85 Should handle state changing in constructors", done => {
...
-                expect(testRoot.getElementsByTagName("span")[0].textContent).toBe(
-                    "child:7 - parent:7"
-                )
-                testRoot.parentNode.removeChild(testRoot)
-                done()
-            }, 10)
-        }, 10)
-    })
+    const { container } = render(<ParentWrapper />)
+    expect(container.getElementsByTagName("span")[0].textContent).toBe("child:3 - parent:3")
```
vkrol: I have to change this assertion from `parent:2` to `parent:3`.

FredyC: Um, why? I mean it does make sense to be like that, but why it would suddenly behave differently?

Also, can you please add `data-testid` attribute to `span` and then use `getByTestId` instead of this verbose thing?

vkrol: > Um, why? I mean it does make sense to be like that, but why it would suddenly behave differently?

It is the side-effect of using `act` from `react-dom/test-utils` in `@testing-library/react` `render`.

> Also, can you please add data-testid attribute to span and then use getByTestId instead of this verbose thing?

 Sure, thanks.

vkrol: > Also, can you please add data-testid attribute to span and then use getByTestId instead of this verbose thing?

I got rid of this query completely. I also added the `@testing-library/jest-dom` package to make the assertions more declarative.

vkrol:
I also started using Jest fake timers in some tests.

vkrol:
@FredyC This part is done. The next part will be done in a few days.

FredyC:
Looking good, thanks again.

PR #733, opened 2019-07-30
Use react-testing-library Part 4

vkrol:
This is the follow up to https://github.com/mobxjs/mobx-react/pull/731.

### DIFF
--- a/test/observer.test.js
+++ b/test/observer.test.js
@@ -1,6 +1,6 @@
 import React, { Component } from "react"
 import { inject, observer, Observer, useStaticRendering } from "../src"
-import { asyncReactDOMRender, createTestRoot, sleepHelper, withConsole } from "./"
+import { withConsole } from "./"
 import renderer, { act } from "react-test-renderer"
 import { render } from "@testing-library/react"
 import * as mobx from "mobx"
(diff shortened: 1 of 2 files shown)


vkrol:
All tests in the repository that used `ReactDOM.render` were rewritten using `react-testing-library`.

test/observer.test.js
```diff
@@ -54,50 +41,68 @@ describe("nestedRendering", async () => {
             }
         }
     )
-    beforeAll(async done => {
-        // the side-effect in  does not views alive when using static rendering test suite
-        useStaticRendering(false)
-        await asyncReactDOMRender(<TodoList />, testRoot)
-        done()
+
+    beforeEach(() => {
```
FredyC: Personally, I find it safer instead of using `beforeEach` to simply call a function in each test and return those variables from that. That way I don't need to rely on closure variables. In my opinion, any shared state makes tests bit more flaky.

vkrol: I agree with you, but I intentionally tried to minimize the surface of the changes. We will be able to refactor these tests in the future, especially since I have a desire to improve something else in them.

PR #737, opened 2019-07-31
Add the ability to pass multiple children to Provider

vkrol:
I think that this is an old restriction when the legacy Context API was used in `Provider`.

### DIFF
--- a/src/Provider.js
+++ b/src/Provider.js
@@ -21,7 +21,7 @@ export class Provider extends Component {
         return createElement(
             MobXProviderContext.Provider,
             { value: this.state },
-            Children.only(this.props.children)
+            this.props.children
         )
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### 6.1.2
 
 -   Add reexport of `useObserver` from `mobx-react-lite` [#734](https://github.com/mobxjs/mobx-react/issues/734)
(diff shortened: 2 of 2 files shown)


src/Provider.js
```diff
@@ -21,7 +21,7 @@ export class Provider extends Component {
         return createElement(
             MobXProviderContext.Provider,
             { value: this.state },
-            Children.only(this.props.children)
+            this.props.children
```
vkrol: I also want to use JSX here, but I will do it in a separate PR.

src/Provider.js
```diff
@@ -21,7 +21,7 @@ export class Provider extends Component {
         return createElement(
```
vkrol: I also want to use JSX here, but I will do it in a separate PR.

FredyC: And you can rewrite it to FC when you are it :D

FredyC:
I suppose it makes sense to lift that restriction. I just checked if `Context.Provider` really does allow multiple children and it's true, so there is no reason to limit that here.

vkrol:
@FredyC Sorry, I forgot to delete `Children` import in this file 🤦‍♂ We need to introduce ESLint to the repository.

src/Provider.js
vkrol: Thanks for the idea :) I'll look into it.

PR #739, opened 2019-07-31
Rewrite the Provider component to function component

vkrol:
From https://github.com/mobxjs/mobx-react/pull/737#discussion_r309406735:
> And you can rewrite it to FC when you are it :D

### DIFF
--- a/src/Provider.js
+++ b/src/Provider.js
@@ -1,39 +1,30 @@
-import { Children, Component, createContext, createElement } from "react"
+import React from "react"
 import { shallowEqual } from "./utils/utils"
 
-export const MobXProviderContext = createContext({})
+export const MobXProviderContext = React.createContext({})
 
-export class Provider extends Component {
-    static contextType = MobXProviderContext
-    static displayName = "MobXProvider"
+export function Provider(props) {
+    const parentValue = React.useContext(MobXProviderContext)
+    const value = React.useRef({
+        ...parentValue,
+        ...grabStores(props)
+    }).current
 
-    constructor(props, context) {
-        super(props, context)
-        this.state = {
-            ...context,
-            ...grabStores(props)
+    if (process.env.NODE_ENV !== "production") {
+        const newValue = { ...value, ...grabStores(props) } // spread in previous state for the context based stores
+        if (!shallowEqual(value, newValue)) {
+            throw new Error(
+                "MobX Provider: The set of provided stores has changed. Please avoid changing stores as the change might not propagate to all children"
+            )
         }
     }
 
-    render() {
-        return createElement(
-            MobXProviderContext.Provider,
-            { value: this.state },
-            this.props.children
-        )
-    }
-
-    static getDerivedStateFromProps(nextProps, prevState) {
-        if (process.env.NODE_ENV !== "production") {
-            const newStores = { ...prevState, ...grabStores(nextProps) } // spread in prevState for the context based stores
-            if (!shallowEqual(prevState, newStores))
-                throw new Error(
-                    "MobX Provider: The set of provided stores has changed. Please avoid changing stores as the change might not propagate to all children"
-                )
-        }
-        return prevState // because they didn't change, remember!
-    }
+    return (
+        <MobXProviderContext.Provider value={value}>{props.children}</MobXProviderContext.Provider>
+    )
 }
+
+Provider.displayName = "MobXProvider"
 
 function grabStores(from) {
     const res = {}


src/Provider.js
```diff
@@ -1,42 +1,37 @@
-import { Children, Component, createContext, createElement } from "react"
+import React, { createContext, useContext, useRef } from "react"
```
FredyC: Can you please stick to the same style of referencing directly through `React.use...`? 

https://github.com/mobxjs/mobx-react/blob/c0ed4b9dcdbe80c277bd03626a89911e2c71fcdb/src/inject.js#L13

src/Provider.js
```diff
@@ -1,42 +1,37 @@
...
-                throw new Error(
-                    "MobX Provider: The set of provided stores has changed. Please avoid changing stores as the change might not propagate to all children"
-                )
+    const value = valueRef.current
+
+    if (process.env.NODE_ENV !== "production") {
+        const newValue = { ...value, ...grabStores(props) } // spread in previous state for the context based stores
+        if (!shallowEqual(value, newValue)) {
+            throw new Error(
+                "MobX Provider: The set of provided stores has changed. Please avoid changing stores as the change might not propagate to all children"
```
FredyC: I am thinking if this is still valid actually. Isn't it from legacy context too? I think it's because the legacy context could not propagate to pure components. The Context does not have that problem. I don't see a reason why the value could not change. I mean in MobX world it doesn't usually need to, but people are crazy sometimes :)

src/Provider.js
```diff
@@ -1,42 +1,37 @@
...
-    static displayName = "MobXProvider"
-
-    constructor(props, context) {
-        super(props, context)
-        this.state = {
-            ...context,
+export function Provider(props) {
+    const parentValue = useContext(MobXProviderContext)
+    const valueRef = useRef(null)
+    if (valueRef.current === null) {
```
FredyC: In light of my other comment, the `useMemo` might work better here, although there is a risk of React throwing it away randomly. So perhaps `useEffect` to update the ref when either parent or value here changes?

src/Provider.js
vkrol: Oh, sure! Miss it.

src/Provider.js
vkrol: The quote from the change log for `mobx-react@6.0.0`:
> * Changing the set of stores in `Provider` is no longer supported and while throw a hard error (this was a warning before), as the model of `Provider` / `inject` has always been to inject final values into the tree. (That is, fixed references, injected objects themselves can be stateful without problem). If you want to dynamically swap what is provided into the tree, use `React.createContext` instead of `Provider` / `inject`. The  suppressChangedStoreWarning` flag for `Provider` has been dropped.

https://github.com/mobxjs/mobx-react/commit/a64c2d365300f669463c0f7d9fcde42337207807#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR11

vkrol: Probably this can be discussed separately from this PR so as not to delay the merge. What do you think?

vkrol: I think we should ask @mweststrate about this change in 6.0.0.

FredyC: I don't think we need a new issue considering this PR is not bug fixing there is no need to hurry.

Presently, I don't see a reason why this should be restricted.

vkrol: I prefer more granular PRs and branches that have one clear goal. It has some connotations with The Single Responsibility Principle, for example. In this PR I want to rewrite the Provider component to function component, so it has not nothing to do with this restriction.
It has the following benefits:
1) the PR will not hung
2) it will be more simple to investigate when and why some changes were made in the future.

FredyC: Well, in that case, would you mind adding some tests for the Provider? If I see correctly, we have none. In theory, this change shouldn't break anything, but better to be sure than sorry. Perhaps even add those tests in a separate PR to see them passing and then merge here.

vkrol: Thanks. Yes, I will look into it.

vkrol: We have some tests in `inject.test.js`. I will move them in a separate file,  review them, refactor them if necessary and probably write new ones.

vkrol: > I don't think we need a new issue considering this PR is not bug fixing so there is no need to hurry.

https://github.com/mobxjs/mobx-react/issues/745

vkrol: @FredyC This PR is ready to merge.

src/Provider.js
```diff
@@ -1,40 +1,35 @@
...
-    static displayName = "MobXProvider"
-
-    constructor(props, context) {
-        super(props, context)
-        this.state = {
-            ...context,
+export function Provider(props) {
+    const parentValue = React.useContext(MobXProviderContext)
+    const valueRef = React.useRef(null)
+    if (valueRef.current === null) {
```
FredyC: Why to actually use that condition here? If you move that init value inside the `useRef` call, the result will be the same imo.

src/Provider.js
```diff
@@ -1,40 +1,35 @@
...
-    }
-
-    static getDerivedStateFromProps(nextProps, prevState) {
-        if (process.env.NODE_ENV !== "production") {
-            const newStores = { ...prevState, ...grabStores(nextProps) } // spread in prevState for the context based stores
-            if (!shallowEqual(prevState, newStores))
-                throw new Error(
-                    "MobX Provider: The set of provided stores has changed. Please avoid changing stores as the change might not propagate to all children"
-                )
+    if (process.env.NODE_ENV !== "production") {
```
FredyC: I would perhaps move this inside the `useEffect` so it doesn't delay the rendering.

src/Provider.js
vkrol: I do not want to create this object on each render. We need to create it once. `React.useRef` does not accept a callback as `React.useState`, so I have to use the condition.

src/Provider.js
vkrol: I think it is good idea to throw as early as possible. Also this condition will delay the rendering in development mode only.

src/Provider.js
FredyC: Well, in that case, the usual trick is `const [value] = React.useState(() => ({ ... }))`.

vkrol: That was the first version of this code. But I intentionally replaced state with ref, because I am sure that the state is not appropriate here, because we will not change them. It is some static data, not mutable state.

FredyC: Haha, look at `mobx-react-lite` hooks, all based on `useState` even though values don't change. Let's do the optimal now, not because it doesn't feel right.

vkrol: Why do you think that `useState` is the optimal way?

vkrol: And in what way is that optimal? What are the criteria?

FredyC: Ok, optimal was a wrong word ... less confusing will be better. It doesn't make immediate sense why you set ref to null, then check for null and set its value. With `useState` you are just creating a one time state and since there is no setter, it's apparent it's a constant value.

FredyC: Or if you have a problem with useState, then really just set value to `useRef`. Note that React is creating so many objects on each render and throwing away. Creating an extra object in Provider which in most cases doesn't even re-render is less than micro-optimization in cost of worse readability.

src/Provider.js
```diff
@@ -1,40 +1,35 @@
...
 
-export class Provider extends Component {
-    static contextType = MobXProviderContext
-    static displayName = "MobXProvider"
-
-    constructor(props, context) {
-        super(props, context)
-        this.state = {
-            ...context,
+export function Provider(props) {
```
FredyC: I am also thinking that `React.memo` could be applied here. Considering props passed here will be some stores with stable reference integrity, it's just fine to prevent that re-render right away. And if someone has primitive value for a _store_ than Provider will re-render just fine (and then fail with that hard exception).

src/Provider.js
vkrol: > Ok, optimal was a wrong word ... less confusing is better. It doesn't make immediate sense why you set ref to null, then check for null and set its value.

This is not something completely new. You can see this idiom in the official Hooks documentation, for example:
https://reactjs.org/docs/hooks-faq.html#how-to-create-expensive-objects-lazily
Other example:
https://github.com/facebook/react/issues/14490#issuecomment-454973512

> Or if you have a problem with useState, then really just set value to useRef. Note that React is creating so many objects on each render and throwing away. Creating an extra object in Provider which in most cases doesn't even re-render is less than micro-optimization in cost of worse readability.

It not about performance only, but about correctness too. We throw an error in the development mode if the stores set changed, but without saving the original value of stores we will rerender with the new set in the production mode and this will be more confusing than rerendering with the original set as we do now.

vkrol: > Or if you have a problem with useState, then really just set value to useRef. Note that React is creating so many objects on each render and throwing away. Creating an extra object in Provider which in most cases doesn't even re-render is less than micro-optimization in cost of worse readability.
 
OK. I will do it. Thanks.

FredyC: > This is not something completely new. You can see this idiom in the official Hooks documentation

Notice the title "How to create **expensive** objects lazily?". We are far from that in here so patterns like that irrelevant, only make code harder to comprehend.

vkrol: > Or if you have a problem with useState, then really just set value to useRef. Note that React is creating so many objects on each render and throwing away. Creating an extra object in Provider which in most cases doesn't even re-render is less than micro-optimization in cost of worse readability.

It is done.

src/Provider.js
FredyC: @vkrol What about this? Not sure why it's marked as resolved...

vkrol: I was thinking that this comment is related to to the comment about getting rid of the stores set changing restriction.

vkrol: I do not think that we should make this change in this PR because it will be definitely a breaking change. Probably we can do it in the future.

FredyC: Yea, it's an actually bad idea ... https://github.com/mobxjs/mobx-react/issues/745#issuecomment-517925519 so nevermind.

vkrol: > No, because Provider depends on children which are always different (unless they're memoized by parent, which is very unlikely). In other words Provider never recieves the same props.

I forgot about this.

vkrol: > No, because Provider depends on children which are always different (unless they're memoized by parent, which is very unlikely). In other words Provider never recieves the same props.

I forgot about this.

FredyC: Me too :)

PR #740, opened 2019-07-31
Use Jest fake timers instead of sleepHelper function

vkrol:


test/hooks.test.js
```diff
@@ -1,103 +1,120 @@
...
+        }
 
-    let wrapper
-    act(() => {
-        wrapper = renderer.create(<Parent />)
-    })
-    expect(wrapper.toJSON()).toMatchInlineSnapshot(`
+        let wrapper
+        act(() => {
+            wrapper = renderer.create(<Parent />)
```
FredyC: Why it's not using `render` from RTL here? Then it wouldn't be necessary to wrap into `act`.

test/hooks.test.js
```diff
@@ -1,103 +1,120 @@
...
-    expect(wrapper.toJSON()).toMatchInlineSnapshot(`
+        jest.runAllTimers()
+
+        expect(seen).toEqual(["parent", 0, "parent", 2])
+        expect(wrapper.toJSON()).toMatchInlineSnapshot(`
 <div>
   2
 </div>
 `)
+    } finally {
```
FredyC: Does it really matter? I mean if a test fails then it's fine that timers are not restored because that dies with that failed Jest process. Or safer alternative can be to restore in `beforeEach`.

test/hooks.test.js
vkrol: That's what I already have in mind. I would like to reconsider the using of `react-test-renderer` in the future PRs.

test/hooks.test.js
vkrol: > Does it really matter? I mean if a test fails then it's fine that timers are not restored because that dies with that failed Jest process.
I think that it is matter because without this if the first test fails, so `jest.useRealTimers` is not called, then the second test will fail too, although it should not. It'll be false negative.
> Or safer alternative can be to restore in `beforeEach`.
I've been thinking about it. I have some doubts about that. If we add a new test into this file in the future which won't need timers, `jest.useRealTimers` will be called anyway 🤔

FredyC: Calling `jest.useRealTimers` while fake ones are not used won't do a harm. It's a noop in that case.

vkrol: I don't have any more convincing arguments, so I'm gonna change that :)

vkrol: Done.

FredyC: Thanks! Much nicer I would say.

vkrol: @FredyC Thank you so much for the review! I hope I'm not boring you yet with my PRs? :)

FredyC: Well, tests are never the best excitement, but I am grateful you don't mind working on that :)

vkrol: Thanks!

PR #742, opened 2019-08-01
Get rid of the suppressChangedStoreWarning flag leftovers

vkrol:
This flag was deleted in 6.0.0 but there are some leftovers. The quote from the 6.0.0 changelog:
> The `suppressChangedStoreWarning` flag for `Provider` has been dropped.

and
> Changing the set of stores in Provider is no longer supported and while throw a hard error (this was a warning before)


FredyC:
Well. Looking at those links it's clear it's related to the legacy context and as such it's misleading.

I am really convinced we shouldn't prevent this if someone really wants that. There is no actual harm in that. It's not about recreating stores, just the context object itself and the only downside is it will re-render a whole tree from that Provider.

vkrol:
I would like to wait for @mweststrate comment about this restriction. I'd like to suggest the following:
1. We are merging this PR (I want to merge this as soon as possible because we have the wrong information about this flag in README.md).
2. I'm creating an issue in which we're waiting for @mweststrate comment.

vkrol:
I want to merge it anyway, because in theory we'll have to wait a long time for comment.

README.md
```diff
@@ -308,7 +308,7 @@ Notes:
...
+-   Values provided through `Provider` should be final, to avoid issues like mentioned in [React #2517](https://github.com/facebook/react/issues/2517) and [React #3973](https://github.com/facebook/react/pull/3973), where optimizations might stop the propagation of new context. Instead, make sure that if you put things in `context` that might change over time, that they are `@observable` or provide some other means to listen to changes, like callbacks. However, if your stores will change over time, like an observable value of another store, MobX will throw an error.
```
FredyC: Let's change it like this at least so there is no misleading information about issues that are not relevant to current implementation anymore.
```suggestion
-   Values provided through `Provider` should be final. Make sure that if you put things in `context` that might change over time, that they are `@observable` or provide some other means to listen to changes, like callbacks. However, if your stores will change over time, like an observable value of another store, MobX will throw an error.
```

vkrol: Sure. Thank you very much for the suggestion.

vkrol:
> I'm creating an issue in which we're waiting for @mweststrate comment.

https://github.com/mobxjs/mobx-react/issues/745

PR #743, opened 2019-08-01
Simplify the special props handling in the Provider component

vkrol:
I think that the special props handling logic in the `Provider` component is partially redundant. `key` and `ref` are special and we do not need to filter them because the props object will never contain them.
> Most props on a JSX element are passed on to the component, however, there are two special props (ref and key) which are used by React, and are thus not forwarded to the component.

https://reactjs.org/warnings/special-props.html


FredyC:
Makes sense, thanks.

vkrol:
No problem.

PR #746, opened 2019-08-02
Restructure Provider component tests

vkrol:
> Well, in that case, would you mind adding some tests for the Provider? If I see correctly, we have none. In theory, this change shouldn't break anything, but better to be sure than sorry. Perhaps even add those tests in a separate PR to see them passing and then merge here.

https://github.com/mobxjs/mobx-react/pull/739#discussion_r309555674

> We have some tests in `inject.test.js`. I will move them in a separate file, review them, refactor them if necessary and probably write new ones.

https://github.com/mobxjs/mobx-react/pull/739#discussion_r309846431


test/Provider.test.js
```diff
@@ -1,6 +1,8 @@
 import React from "react"
-import { Provider } from "../src"
+import { observable } from "mobx"
+import { observer, Provider } from "../src"
 import { render } from "@testing-library/react"
+import renderer, { act } from "react-test-renderer"
```
FredyC: I don't follow why to use this one when line above you have RTL. Let's not mix two approaches in a single test file please :)

vkrol: It is already done :) I will do this step by step.

vkrol: UPD: This is the final version of the PR.

FredyC:
Looks great, thanks!

vkrol:
@FredyC thank you!

PR #747, opened 2019-08-02
Check shallowly for state changes on sCU

liady:
Check shallowly for state changes on sCU (and not just referential equality)
In order to be aligned with React's PureComponent, which checks shallowly for props and state:
```
    } else if (
      elementType.prototype &&
      elementType.prototype.isPureReactComponent
    ) {
      shouldUpdate =
        !shallowEqual(oldProps, props) || !shallowEqual(oldState, state);
    }
```

### DIFF
--- a/src/observerClass.js
+++ b/src/observerClass.js
@@ -114,15 +114,11 @@ function observerSCU(nextProps, nextState) {
             "[mobx-react] It seems that a re-rendering of a React component is triggered while in static (server-side) mode. Please make sure components are rendered only once server-side."
         )
     }
-    // update on any state changes (as is the default)
-    if (this.state !== nextState) {
-        return true
-    }
     // update if props are shallowly not equal, inspired by PureRenderMixin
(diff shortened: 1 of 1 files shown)


FredyC:
I am pretty sure this is intentional and it broke the test which explicitly checks for that. I don't have any explanation why is it like that, so I am leaving it open for a further discussion.

liady:
I see, and this will probably be a breaking change as well. In my opinion this (the new behavior) is the expected behavior - but if there is indeed a good reason behind the current behavior, we should maybe leave it as it is.

mweststrate:
The reason is that `observer` makes the thing pure in terms of props (which is important, because that is how we typically get by our mutable observable objects), but not in terms of `state`, as it doesn't interact with `state` at all (nor is there much reason to use `state` in `observer` components in the first place. So for `state` we stick to the default behavior, and if you want to optimize that, I'd recommend to use local observables instead of state :)

PR #748, opened 2019-08-03
Use jest-mock-console instead of the withConsole util

vkrol:
@FredyC 
> Can we perhaps use `jest-mock-console` instead?

https://github.com/mobxjs/mobx-react/pull/724#discussion_r302284153

@vkrol 
> It has nothing to do with `react-testing-library`. I can investigate it in the separate PR, if you don't mind.

https://github.com/mobxjs/mobx-react/pull/724#discussion_r302381652


test/context.test.js
```diff
@@ -34,17 +36,16 @@ test("no warnings in modern react", () => {
...
-    ).toEqual({ errors: [], infos: [], warnings: [] })
+    const testRenderer = TestRenderer.create(<App />)
+    expect(testRenderer.toJSON()).toMatchSnapshot()
+
+    TestRenderer.act(() => {
+        box.set(4)
+    })
+    expect(testRenderer.toJSON()).toMatchSnapshot()
+
+    expect(console.warn).not.toHaveBeenCalled()
```
FredyC: Originally it was checking for errors and infos as well. Shouldn't be here as well just to be sure.

test/inject.test.js
```diff
@@ -154,11 +156,10 @@ describe("inject based context", () => {
                 )
             }
         }
-        withConsole(() => {
-            expect(() => renderer.create(<A />)).toThrow(
-                /Store 'foo' is not available! Make sure it is provided by some Provider/
-            )
-        })
+        expect(() => renderer.create(<A />)).toThrow(
```
FredyC: Generally, I try to limit the area where I mock out console in test because it can also swallow other things. Especially painful when trying to console.log something and wondering why it's not there :) That's what `withConsole` was good for. Perhaps we can keep that wrapper but use jest variant inside?

test/observer.test.js
```diff
@@ -387,6 +387,8 @@ describe("124 - react to changes in this.props via computed", () => {
 // Test on skip: since all reactions are now run in batched updates, the original issues can no longer be reproduced
 //this test case should be deprecated?
 test("should stop updating if error was thrown in render (#134)", () => {
+    mockConsole()
```
FredyC: Why no restore?

test/__snapshots__/context.test.js.snap
```diff
@@ -13,23 +13,3 @@ exports[`no double runs for getDerivedStateFromProps 1`] = `
   Test-content
 </div>
 `;
-
```
FredyC: Why remove these snapshots? Tests are failing because of that.

vkrol: It is very strange 🤔

test/observer.test.js
vkrol: By mistake. Thanks!

test/inject.test.js
vkrol: What do you think about moving the `mockConsole` invocation right before this expect?

test/context.test.js
vkrol: I agree, thanks.

test/inject.test.js
vkrol: I've moved the `mockConsole` invocations as close as I can to where they are needed.

FredyC: Why not keep that `withConsole` as I drafted? It's very handy as you don't need to remember to restore...

```js
function withConsole(cb => {
  const restoreConsole = mockConsole()
  cb()
  restoreConsole()
})
```

vkrol: I'm not sure yet that introducing additional abstraction around the jest-mock-console is worth it, but I'll do it because I'm not 100% sure.

FredyC: It's just for convenience, I don't exactly insist, it just looks safer that way.

vkrol: OK. Thanks. Then it is the final version of PR.

test/context.test.js
```diff
@@ -1,8 +1,8 @@
 import React from "react"
 import * as mobx from "mobx"
 import { Provider, observer, inject } from "../src"
-import { withConsole } from "./"
 import TestRenderer from "react-test-renderer"
+import { withConsole } from "."
```
FredyC: Um, can we actually have it in a `withConsole.ts` file? These imports from the index are strange :)

vkrol: I will create `test/utils/withConsole.js` module, if you do not mind.

FredyC: Sure, that works as well :)

vkrol: It's done.

FredyC: Almost...check tests failure :)

vkrol: I'm sorry. I forgot to add some changes to Git 🤦🏻‍♂️

FredyC:
Looking good, thanks!

PR #749, opened 2019-08-04
Setup missing React act(...) warning workaround globally in tests

vkrol:
The advantage is that we do not need to import `test/index.js` manually in each test.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,0 +1,8 @@
+// Uglyness to find missing 'act' more easily
+// 14-2-19 / React 16.8.1, temporarily work around, as error message misses a stack-trace
+Error.stackTraceLimit = Infinity
+const origError = console.error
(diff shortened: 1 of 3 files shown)


FredyC:
Well, in the essence the `act` should not be needed to use at all with latest RTL and React 16.9 alpha so I would rather go that way instead of hacks.

vkrol:
For now this hack does not work properly because we do not import `test/index.js` in all tests, so I think it is better to merge this PR. I think that using the alpha version of React is not a good idea because it can introduce some "flakiness" to the tests because is is alpha, not stable version.

FredyC:
I suppose it's safe to merge this event we don't upgrade React eventually. Although, there are clearly no missed `act` calls :)

vkrol:
@FredyC I agree. Thank you!

PR #751, opened 2019-08-04
Introduce ESLint

vkrol:


.eslintrc.js
```diff
@@ -0,0 +1,39 @@
...
+        process: "readonly"
+    },
+    plugins: ["react"],
+    rules: {
+        // Conflicted with Prettier
+        "no-extra-semi": "off",
+        "react/jsx-uses-react": "error",
+        "react/jsx-uses-vars": "error"
+    },
+    overrides: [
```
FredyC: Let's keep this contained to specific places. For a file, you can just add comment `/* eslint-env node */`. For tests create another `.eslintrc.js` which overrides env (see https://eslint.org/docs/user-guide/configuring#using-configuration-files-1).

package.json
```diff
@@ -16,6 +16,7 @@
   },
   "scripts": {
     "prettier": "prettier --write \"**/*.js\" \"**/*.ts\"",
+    "lint": "eslint .",
     "test": "jest && yarn test:ts",
     "test:ts": "tsc -p test/ts",
     "test:ci": "jest -i --coverage && yarn test:ts",
```
FredyC: Add `lint` here as well, after tests should be better.

test/Provider.test.js
```diff
@@ -24,7 +24,7 @@ describe("Provider", () => {
                 <Provider>
                     <MobXProviderContext.Consumer>
                         {stores =>
-                            stores.hasOwnProperty("children")
+                            Object.prototype.hasOwnProperty.call(stores, "children")
```
FredyC: Let's use [`Reflect.has`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Reflect/has) instead.

vkrol:
This is the final version of this PR. I am not sure about adding TypeScript support to the ESLint configuration for now. I think we can do this later.

FredyC:
Thanks, this will help a lot. I agree we don't need TypeScript since this project is still written in JS. I would like to eventually tackle that.

vkrol:
OK, thanks!

PR #752, opened 2019-08-05
Use the react-testing-library rerender function instead of render to the container

vkrol:
This is the small PR with mostly cosmetic changes. I didn't know about this function when I refactored these tests.


FredyC:
Looking good. Are you planning to get rid of react-test-renderer completely? I thought you mentioned it somewhere.

vkrol:
I'm thinking about it, but it's gonna take some time.

PR #762, opened 2019-08-14
Use same naming scheme and styles in imports in tests

vkrol:
When working with the tests, I constantly encountered different naming and import styles in the tests. I think it's best to use the same naming style for imports. This reduces the cognitive load when working with the existing code and allows you not to think about what style to use when writing a new code.

I also want to use the same style in modules located in the `src directory`. If you don't mind, I'll do it in a separate PR.

### DIFF
--- a/test/context.test.js
+++ b/test/context.test.js
@@ -1,11 +1,11 @@
 import React from "react"
-import * as mobx from "mobx"
+import { observable } from "mobx"
 import { Provider, observer, inject } from "../src"
 import TestRenderer from "react-test-renderer"
 import withConsole from "./utils/withConsole"
 
 test("no warnings in modern react", () => {
-    const box = mobx.observable.box(3)
+    const box = observable.box(3)
     const Child = inject("store")(
         observer(
             class Child extends React.Component {
--- a/test/disposeOnUnmount.test.js
+++ b/test/disposeOnUnmount.test.js
@@ -1,4 +1,4 @@
-import * as React from "react"
+import React from "react"
 import { disposeOnUnmount, observer } from "../src"
 import { render } from "@testing-library/react"
 
--- a/test/hooks.test.js
+++ b/test/hooks.test.js
@@ -1,6 +1,6 @@
-import React, { useState, useEffect } from "react"
+import React from "react"
 import { observer, Observer, useLocalStore, useAsObservableSource } from "../src"
-import renderer, { act } from "react-test-renderer"
+import TestRenderer, { act } from "react-test-renderer"
 
 afterEach(() => {
     jest.useRealTimers()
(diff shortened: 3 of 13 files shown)


test/context.test.js
```diff
@@ -1,11 +1,11 @@
 import React from "react"
-import * as mobx from "mobx"
+import { observable } from "mobx"
```
FredyC: To be honest, I am more inclined to another version. Consider that we are doing `React.useState` so it should be the same `mobx` to improve consistency. Sure, it's some extra typing, but at least it's the origin of that call is clear.

vkrol:
> I wonder if there are some ESLint rules to enforce that. Otherwise, I agree to do it for src as well.

I don't know any.

test/context.test.js
vkrol: Generally I don't think it's necessary to import all the modules in the same style. It is more important to import the same module in different files in the same way.
I don't see any point in using a namespace for mobx because we are "in the context of MobX". I used a different style for React because I relied on that comment https://github.com/mobxjs/mobx-react/pull/739#discussion_r309443755 but now I think it was a mistake and it's better to use named imports for React too. Therefore, I suggest using named imports for both React and MobX.

vkrol: I don't want to bother you, but I'd like to know what I need to do to get things moving.

FredyC: There is a good reason for `React.useState` style ... to distinguish from custom hooks by the glance. With only `useState` you have to constantly be checking if it's from React or what is it.

It's true there is not a similar issue with mobx, but it still improves code readability when you know the namespace of that function. Otherwise, it could be a local function, argument or anything. It's just my opinion of course. If you don't agree, I guess you have to lure someone else here to decide.

vkrol: > There is a good reason for React.useState style ... to distinguish from custom hooks by the glance. With only useState you have to constantly be checking if it's from React or what is it. 

I do no think that any library will have a hook with `useState` name.

> Many popular libraries have already adopted that, eg. Formik. It's not something from my mind, but I do like it and use it regularly everywhere.

OK. What about using different styles for React and MobX? In my opinion, importing MobX as namespace is much less common than importing it via named imports. For example:
https://github.com/mobxjs/mobx-react/blob/master/src/index.js#L1
https://github.com/mobxjs/mobx-state-tree/blob/master/packages/mobx-state-tree/src/index.ts#L2
https://github.com/mobxjs/mobx-react-lite/blob/master/src/useLocalStore.ts#L1
Also:
> I don't see any point in using a namespace for MobX because we are "in the context of MobX".

vkrol:
@FredyC thank you very much for code reviews and interesting discussions!

PR #766, opened 2019-08-24
Use react-testing-library instead of react-test-renderer Part 1

vkrol:


vkrol:
I don't want to bother you, just in case, if these is no notification.

vkrol:
> Yea, there wasn't any :) I mean there was a notification when you created a draft and then with additional changes, but never got one that it's ready for review. Silly GitHub :)

:( OK, then I will send a message when the PR is ready to merge.

> Looking good, thanks again for the great work.

Thank you!

PR #768, opened 2019-08-27
Use react-testing-library instead of react-test-renderer part 2

vkrol:
This is the follow-up to https://github.com/mobxjs/mobx-react/pull/766.

### DIFF
--- a/test/misc.test.js
+++ b/test/misc.test.js
@@ -1,7 +1,6 @@
 import React from "react"
-import { extendObservable, isObservable, observable } from "mobx"
+import { extendObservable, observable } from "mobx"
 import { observer } from "../src"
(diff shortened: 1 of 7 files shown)


test/misc.test.js
```diff
@@ -86,15 +86,6 @@ test("#85 Should handle state changing in constructors", () => {
     expect(container).toHaveTextContent("child:7 - parent:7")
 })
 
-test("testIsComponentReactive", () => {
```
vkrol: I intentionally removed this test. I don't see any sense in it.

vkrol:
Thank you. Also I would like to simplify and rewrite some tests in the future because some tests are very complex and have unnecessary details.

PR #781, opened 2019-09-22
fix PureComponent class check for length > 1 prototype chains

aspectgfg:
The __proto__ check fails if a component doesn't directly extend from `PureComponent`. Example:

```js
@observer
class ItemCell extends PureComponent {
}

@observer
class MessageCell extends ItemCell {

}
```

Results in:
```
Warning: MessageCell has a method called shouldComponentUpdate(). shouldComponentUpdate should not be used when extending React.PureComponent. Please extend React.Component if shouldComponentUpdate is used.
```

Changing the inherits from PureComponent check to be `componentClass.prototype instanceof PureComponent` takes care of the problem.

### DIFF
--- a/src/observerClass.js
+++ b/src/observerClass.js
@@ -13,7 +13,7 @@ export function makeClassComponentObserver(componentClass) {
     const target = componentClass.prototype
     if (target.componentWillReact)
         throw new Error("The componentWillReact life-cycle event is no longer supported")
-    if (componentClass.__proto__ !== PureComponent) {
+    if (!(target instanceof PureComponent)) {
         if (!target.shouldComponentUpdate) target.shouldComponentUpdate = observerSCU
         else if (target.shouldComponentUpdate !== observerSCU)
             // n.b. unequal check, instead of existence check, as @observer might be on superclass as well


vkrol:
I think that we need a test for that.

FredyC:
It's kinda discouraged to use inheritance with React components. You should use HOC instead really or convert to Hooks. I don't think we want to openly support such anti-pattern here.

kidcosmic:
I agree Hooks are better, but the issue here isn't "are we supporting classes". It's "we're incorrectly supporting classes". As for tests, I'd probably need a day or two to swing around to it.

FredyC:
It's merely a question of supporting React anti-pattern and letting people shot themselves in the foot. But I guess too many restrictions can be annoying as well.

Just keep in mind you would be better off with HOC than the inheritance of components. You can find the official recommendation in [React docs](https://reactjs.org/docs/composition-vs-inheritance.html) plus some reasoning behind it.

JunlinZhu-Tommy:
weird, why we didn't use instance of before?

JunlinZhu-Tommy:
Why don't we merge this?

FredyC:
I have given some reasoning, waiting for others to chime in because I don't think this should be supported even though it's trivial change. You should be abusing inheritance in React.

kidcosmic:
The point of this repo though is class support isn't it? mobx-react-lite
for hooks?

FredyC:
Nope, `mobx-react` is universal and includes `mobx-react-lite`

That's not the point though. React class components should not be inherited. Use HOC if you need to share functionality.

mweststrate:
> @mweststrate Do you recall why not instanceof?

I don't think there was a specific reason for it. 

The reason we don't support inheritance trees is because it has to much footguns: It only works if exactly _only_ the finally instantiated class(es) are marked as `observer`. Which is an complicated rule. But if this solves the use case I think it is fine if tests are added. The check is more idiomatic anyway.

mweststrate:
Short reminder that this PR is still pending tests :)

FredyC:
I will close this for now. Anyone feel free to make a new PR since it's such a trivial change, but please include tests.

dperetti:
> React class components should not be inherited. 
Tsssss! It's perfectly fine to subclass a component. 

> Use HOC if you need to share functionality.
Well, no thanks. :-) We may have a perfectly valid reason to use inheritance!

It reminds me of advice from react team members: "don't use mobx, setState() is the recommended way to manage the state" 🙂

mweststrate:
@dperetti this PR was primarily closed because nobody needing this feature was willing to put effort in it.

PR #787, opened 2019-10-15
Opt-in variant of batched updates

FredyC:
Mostly just reexport of the https://github.com/mobxjs/mobx-react-lite/pull/214.

In some distant future when 7.0 comes out, we can remove react-native specific bundle. Although it would be nice to somehow deprecate it now. Ideas?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ There are currently two actively maintained versions of mobx-react:
 | v6          | 16.8.0 and higher        | Yes                                                                              |
 | v5          | 0.13 and higher          | No, but it is possible to use `<Observer>` sections inside hook based components |
 
-The user guide covers this [in a more detail](https://mobx-react.js.org/libraries). 
+The user guide covers this [in a more detail](https://mobx-react.js.org/libraries).
 
 The V5 documentation can be found in the [README_v5](README_v5.md).
 
@@ -319,6 +319,7 @@ Notes:
 -   The original component wrapped by `inject` is available as the `wrappedComponent` property of the created higher order component.
 
 #### "The set of provided stores has changed" error
+
 Values provided through `Provider` should be final. Make sure that if you put things in `context` that might change over time, that they are `@observable` or provide some other means to listen to changes, like callbacks. However, if your stores will change over time, like an observable value of another store, MobX will throw an error.
 This restriction exists mainly for legacy reasons. If you have a scenario where you need to modify the set of stores, please leave a comment about it in this issue https://github.com/mobxjs/mobx-react/issues/745. Or a preferred way is to [use React Context](https://mobx-react.js.org/recipes-context) directly which does not have this restriction.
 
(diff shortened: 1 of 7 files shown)


FredyC:
Well, and what do you think about deprecation strategy for the current approach?

mweststrate:
I think this is a cool approach, alleviating some tooling issues! As a hint, we could maybe detect and warn during the first render that no scheduler has been set at all. (and maybe even provide a noop scheduler if for some reason observer is used without react-dom or native)

FredyC:
I am still a bit concerned about the deprecation strategy. Or do you think that removing RN specific bundle with a next major without any sort of warning is ok? It's probably not that big change on the consumer side, that's true.


> As a hint, we could maybe detect and warn during the first render that no scheduler has been set at all.

Wouldn't that be kinda obtrusive? Most people probably don't need it and bugging them to even set noop one feels like a nuisance.

mweststrate:
imho most people does need it at some point, as not using this will lead at
random to really confusion semantics / /bugs when an update causes both a
parent and child should update, and they don't do it in order. So I quite
disagree with the premise ("most people don't need it"). It is imho simply
a requirement for mobx-react to run correctly and predictable. The fact
that it doesn't cause a bug in all projects (or even in most, if you stick
to causing updates from event handlers), doesn't mean it's a random opt-in
optimization; its not just an optimization, its a change in semantics.

Figuring out afterwards why something doesn't work here, and why components
are updating in the incorrect order, is way way harder, than blindly
configuring this correctly upfront.

FredyC:
Alright, I seem to have misunderstood the problem at hand. I even call it optimization [in mobx-react-lite docs](https://github.com/mobxjs/mobx-react-lite#optimize-rendering).

It would be nice if you can create some small writeup of the implication of (not) using it. We could then link it from that hint so people can make educated decision and not just a guess.

mweststrate:
Short summary (I think you can confirm this by disabling it in the mobx-react repo and running the unit tests, some should break). For example this test: https://github.com/mobxjs/mobx-react/blob/master/test/observer.test.js#L639

The idea is: 

1. Render two observer based components `<Parent><Child /></Parent>` (btw, you can also reproduce with normal react state
2. Make sure both parent and child depend on the same observable (in the test: `store.user`)
3. Make sure that child has some implicit dependency on the parent (in the test: child wouldn't be rendered by parent if `store.user` is not set)
4. Update that shared observable. E.g. set `store.user` to `undefined`. Now both components will signal to update. However, it is imperative that the Parent is updated before the Child. As the Parent will then remove the child and make sure it never gets updated as all. If the Child would render first, it would try to pick the `name` property of `store.user`. Which would correctly throw; the Child shouldn't even try to render if the user is not set, the Parent would have prevented that if it renders first.
5. Normally, React guarantees that Parents _always_ render before Childs. It does this by using event handlers as batch handlers (very similar to action in MobX). So at the end of an event handler, all components scheduled for update are ordered correctly by react, and then they are actually rendered. This guarantees in most situations that Parents render before Child.
6. However, if an update does not originate from a React event handler, this batching never happens, and this guarantee disappears. For example if you put an timeout in an event handler, or update from a fetch callback, websocket message, etc etc. They might change this in a future of React, but to make sure those cases don't break now, we need to batch manually using `unstable_batchedUpdates` (all similar libraries basically need to do this). Luckily, since mobx has already a batching mechanism we can just combine the two, which is what this whole mechanism is doing; attaching reacts updates to our own 'event loop'. Note that this is also the reason `act` is needed in React unit tests. 
7. So, this issue will not occur in all projects, as you need a combination of some Parent + child relationship + non-react event handlers (and even then, without batching the undeterministic order can be by chance be correct). But as you can see, without any background knowledge, it will be very tricky to figure out what goes wrong if it goes wrong and this wasn't configured by default.

Hope that helps!

FredyC:
Thank you for the write-up, I need to dive into it and try it out and then I will add explanation in some form to mobx-react-docz site.

However, I am thinking there might be a small flaw to this solution - UMD. Since it's not an optimization as I first thought, it's probably viable to have that included in UMD build which would not have an option to do it. But I wonder if people using UMD actually do need it. I cannot think of reasons why would someone use UMD for serious app development. What do you think?

PR #798, opened 2019-11-12
Fix for component-class not disposing reaction if render() is hot-swapped

### LINKED ISSUE #797: Component-class with @observer doesn't dispose reaction if render() is hot-swapped
[mobx-react version: 6.1.4]

For a React component-class with the `@observer` decorator, instances of the class do not dispose their MobX reactions, if their `render()` method is overridden.

For example:
```
class MyButton extends Component {
	render() {
		return <button>MyButton</button>;
	}
	render_withLogging() {
		console.log("Start");
		return <button>MyButton</button>;
		console.log("End");
	}
	componentDidMount() {
		if (window.enableLogging) {
			this.render = this.render_withLogging;
		}
	}
}
```

Since the component above hot-swaps the render function, the `this.render[mobxAdminProperty]` attachment becomes inaccessible to the following code in mobx-react's patched version of `componentWillUnmount()`: https://github.com/mobxjs/mobx-react/blob/a4099e86f07b3c0778da92663a82746c077c1756/src/observerClass.js#L36-L40

Become the reaction (stored under the `mobxAdminProperty` key) is inaccessible, the call to `[reaction].dispose()` is short-circuited, meaning it never gets cleaned up.

For reference, here is where that special property is first attached: (it only runs the first time the component is rendered, so its hot-swapping of the render function is not guaranteed to be final) https://github.com/mobxjs/mobx-react/blob/a4099e86f07b3c0778da92663a82746c077c1756/src/observerClass.js#L88-L89

### Real-world library conflicts

The reason I discovered the issue is because I've been using this library to enable hooks in react class-components: https://github.com/salvoravida/react-class-hooks

It conflicts because, as alluded to above, the library hot-swaps the render function of component instances just like mobx-react does, except it does it even later than mobx-react. (when a hook is first used, rather than when the render function merely begins)

### Potential solutions

1) Attach the reaction (ie. value under `mobxAdminProperty`) to the class instance rather than the render function.
2) Patch the `this.render` to be a getter+setter rather than a property, and code the setter such that, if the value gets replaced, the special `mobxAdminProperty` property (ie. the reaction) gets transferred from the old render-function to the new one.

I favor the first solution since it doesn't add another level of indirection.

Thoughts?

Venryx:
* Fixed issue 797 (reaction never being disposed if this.render gets hot-swapped) by having the reaction be stored on the component instance itself, rather than its render function. (since its render function may be swapped out, eg. wrapping the render-func to add logging) Resolves #797.
* Added test for the above issue, to prevent regressions.

### DIFF
--- a/src/observerClass.js
+++ b/src/observerClass.js
@@ -1,10 +1,11 @@
 import { PureComponent, Component } from "react"
-import { createAtom, _allowStateChanges, Reaction, $mobx } from "mobx"
+import { createAtom, _allowStateChanges, Reaction } from "mobx"
 import { isUsingStaticRendering } from "mobx-react-lite"
 
 import { newSymbol, shallowEqual, setHiddenProp, patch } from "./utils/utils"
 
-const mobxAdminProperty = $mobx || "$mobx"
+export const mobxAdminPropertyForRender =
+    typeof Symbol !== "undefined" ? Symbol("mobx administration - render") : "$mobx_render"
 const mobxIsUnmounted = newSymbol("isUnmounted")
 const skipRenderKey = newSymbol("skipRender")
 const isForcingUpdateKey = newSymbol("isForcingUpdate")
@@ -35,7 +36,7 @@ export function makeClassComponentObserver(componentClass) {
     }
     patch(target, "componentWillUnmount", function() {
         if (isUsingStaticRendering() === true) return
-        this.render[mobxAdminProperty] && this.render[mobxAdminProperty].dispose()
+        this[mobxAdminPropertyForRender] && this[mobxAdminPropertyForRender].dispose()
         this[mobxIsUnmounted] = true
     })
     return componentClass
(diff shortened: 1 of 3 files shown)


src/observerClass.js
```diff
@@ -35,7 +35,7 @@ export function makeClassComponentObserver(componentClass) {
     }
     patch(target, "componentWillUnmount", function() {
         if (isUsingStaticRendering() === true) return
-        this.render[mobxAdminProperty] && this.render[mobxAdminProperty].dispose()
+        this[mobxAdminKeyForRender] && this[mobxAdminKeyForRender].dispose()
```
FredyC: What if we actually keep in `render` as well for compatibility purposes, but disposal will happen based on what value from the instance?

Venryx:
Closing since the solution in this PR is incomplete. (if something modifies compInstance.render, this PR fixes the memory leak, but it doesn't solve the other problem of MobX not having a reference to the new function, and therefore being unable to trigger it if an observed value changes)

PR #799, opened 2019-11-12
Warn if attempting to modify comp.render after MobX has attached

Venryx:
* Added warning for attempting to modify a component class' render function after a MobX reaction has already attached. Helps prevent memory leaks as in #797. (So mostly resolves [#797](https://github.com/mobxjs/mobx-react/issues/797))
* Added test for the above.
* Added changelog entry.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # MobX-React Changelog
 
+### 6.1.5
+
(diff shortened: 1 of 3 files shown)


FredyC:
Please, make it DEV only check.

Venryx:
> Please, make it DEV only check.
Done.

mweststrate:
Great stuff, thanks!

PR #821, opened 2020-01-23
Use TSDX for bundling

### LINKED ISSUE #820: Revert #812
I believe we need to revert this change: #812 
1. if `process` is not available in the global scope (and it is not in browsers) it will break because you can't check for variable existence by doing it like
```
if (nonExistentVar) { ... }
```
2. It breaks dead-code elimination for some reason, probably because of a more complex expression.

FredyC:
I've suggested it in the past and now when conversion to TS is happening, this is a good opportunity.

Names of distribution files have changed. It matters for UMD only and I made a simple file copy so it's not a breaking change for now.

Fixes #820


src/Provider.tsx
```diff
@@ -1,17 +1,21 @@
...
+    const { children, ...stores } = props
     const parentValue = React.useContext(MobXProviderContext)
-    const value = React.useRef({
-        ...parentValue,
-        ...stores
-    }).current
+    const mutableProviderRef = React.useRef({ ...parentValue, ...stores })
+    const value = mutableProviderRef.current
 
-    if (process.env.NODE_ENV !== "production") {
```
ynejati: You probably already know this already, but just in case so you don't forget, you'll have to do the same thing on line 42 in `observerClass`.

FredyC: Yea, I am aware, thanks :)

FredyC:
> I am noticing that we aren't building development and production version of the ES modules. Was that intentional? See screenshot below of modules in the distribution folder.

Yes, that's intentional because you cannot run ES anywhere without bundling it which means bundler will take care of creating appropriate version based on env.

PR #823, opened 2020-01-26
Issue #822: resolve signature of observer decorator

### LINKED ISSUE #822: TypeScript unable to resolve signature of observer decorator
I failed to mention yesterday that I wasn't sure that [818](https://github.com/mobxjs/mobx-react/pull/818) was entirely ready.  While converting the tests to TypeScript, I ran across this error with the observer decorator.  To prevent a regression, we should revert back to the old shape, which appears to be correct.  I wouldn't be surprised if I run into similar problems as a work my way through these tests.  I will submit a PR soon, stand by.

```
Unable to resolve signature of class decorator when called as an expression.
  Type 'ClassicComponentClass<any>' is not assignable to type 'void | typeof Component'.
    Type 'ClassicComponentClass<any>' is not assignable to type 'typeof Component'.
      Property 'swapRenderFunc' is missing in type 'ClassicComponent<any, any>' but required in type 'Component'.ts(1238)
```

ynejati:
Resolves #822.  Fallout from JavaScript to TypeScript conversion.  I suppose I may run into more issues as a work my way through the tests.  I will have a better idea by the end of the day.

### DIFF
--- a/src/observer.tsx
+++ b/src/observer.tsx
@@ -18,7 +18,7 @@ const ReactMemoSymbol = hasSymbol
 /**
  * Observer function / decorator
  */
-export function observer(component: IReactComponent<any>): IReactComponent<any> {
+export function observer<T extends IReactComponent>(component: T): T {
     if (component["isMobxInjector"] === true) {
         console.warn(
             "Mobx observer: You are trying to use 'observer' on a component that already has 'inject'. Please apply 'observer' before applying 'inject'"
@@ -40,7 +40,7 @@ export function observer(component: IReactComponent<any>): IReactComponent<any> 
             throw new Error("render property of ForwardRef was not a function")
         return React.forwardRef(function ObserverForwardRef() {
             return <Observer>{() => baseRender.apply(undefined, arguments)}</Observer>
-        })
+        }) as any
     }
 
     // Function component
@@ -50,8 +50,8 @@ export function observer(component: IReactComponent<any>): IReactComponent<any> 
         !component["isReactClass"] &&
         !Object.prototype.isPrototypeOf.call(React.Component, component)
     ) {
-        return observerLite(component as React.StatelessComponent<any>)
+        return observerLite(component as React.StatelessComponent<any>) as T
     }
 
-    return makeClassComponentObserver(component as React.ComponentClass<any, any>)
+    return makeClassComponentObserver(component as React.ComponentClass<any, any>) as T
 }
--- a/src/observer.tsx
+++ b/src/observer.tsx
@@ -40,7 +40,7 @@ export function observer<T extends IReactComponent>(component: T): T {
             throw new Error("render property of ForwardRef was not a function")
         return React.forwardRef(function ObserverForwardRef() {
             return <Observer>{() => baseRender.apply(undefined, arguments)}</Observer>
-        }) as any
+        }) as T
     }
 
     // Function component
--- a/src/propTypes.ts
+++ b/src/propTypes.ts
@@ -1,7 +1,7 @@
 import { isObservableArray, isObservableObject, isObservableMap, untracked } from "mobx"
 
 // Copied from React.PropTypes
-function createChainableTypeChecker(validator: React.Validator<any>): React.Validator<any> {
+function createChainableTypeChecker(validator: React.Validator<any>): React.Requireable<any> {
     function checkType(
         isRequired: boolean,
         props: any,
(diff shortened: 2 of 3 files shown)


src/observer.tsx
```diff
@@ -40,7 +40,7 @@ export function observer(component: IReactComponent<any>): IReactComponent<any>
             throw new Error("render property of ForwardRef was not a function")
         return React.forwardRef(function ObserverForwardRef() {
             return <Observer>{() => baseRender.apply(undefined, arguments)}</Observer>
-        })
+        }) as any
```
FredyC: Perhaps we might include `RefForwardingComponent` in `IReactComponent` as well? Given that we actually try to identify such a component, it should be supported to be passed in.

ynejati: Sounds good. Also replaced Stateless with Functional in next commit.

src/observer.tsx
```diff
@@ -40,7 +40,7 @@ export function observer<T extends IReactComponent>(component: T): T {
             throw new Error("render property of ForwardRef was not a function")
         return React.forwardRef(function ObserverForwardRef() {
             return <Observer>{() => baseRender.apply(undefined, arguments)}</Observer>
-        }) as any
+        }) as T
```
ynejati: I am unsure of the casting and wonder if there is a better way.  Open to ideas.

FredyC: Not sure either, I would be expecting it would work directly as those types are compatible. Unless that generic argument for props is a problem somehow. The `React.forwardRef<R, P>` does probably generate the same output.  Perhaps you can try this and then utilize the `P` there 🤷‍♂ 

`observer<P extends object, T extends IReactComponent<P>>(component: T): T {`

mweststrate: I think for observer we should keep exactly the same signature as we had before in the `.d.ts`, two not be confused with two changes at the same time.... That means the old TS types should remain working out of the box?

FredyC: Well, I think it should be just fine. Since the `P` is referenced in the `T`, it won't be required to be used. I highly doubt anyone has been supplying `T` with previous versions.

https://www.typescriptlang.org/play/index.html#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wChSkAPSWOGATzCTgEll0YBhXSAOyV4wAPAAU4AXjgpe9AHwTScJXAA+iYjAB0nADYoAzvuBpu4CP0G6D+0bMXK17DNp7mBXPYdv2ljjZoBlGBQYJB0kQ1M+d29lVXUOTWRMADFoAHcUKAATYF4Acyi3QW9STABXXgxgczgIACN9JCgAN2bROCpQ3mz9OvqAKyQMABo4ABVOym7e1icuVwthEVlZAAo0RfcALgmASl3JgG8fOCIYcqheOE2zJdIAX3IGptbmtbW9iXlecp0dPZAA

On the contrary, with this updated variant we are aligning with `observer` from `mobx-react-lite` which does accept `P` as generic arg.

ynejati: Yeah, I tried your suggestion and I still get the same error.  Where required to cast to T the return value of  `forwardRef` (ExoticComponent), `observerLite` (FunctionComponent), and `makeClassComponentObserver` (ComponentClass).  In order to avoid

`Type is not assignable to type 'T'.
  Type is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'IReactComponent<P>'`

FredyC: Well, and did you try `React.forwardRef<any, P>`?

ynejati: Yes, I believe so.  Like this?

`observer<P extends object, T extends IReactComponent<P>>(component: T): T`

`React.forwardRef<any, P>(function ObserverForwardRef() {`

Output:
`Type 'ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<any>>' is not assignable to type 'T' `

ynejati: I guess all in all I'm thinking we keep the signature the same for now so that there are no regressions.  Then, as I work through the tests, I will also spend some time coming up with a better solution (if any exists).  I wouldn't want this to hold up a potential release.

FredyC: Alright, fair enough, let's keep it simple for now and feel free to look into it over the time.

ynejati:
A quick glance, it looks like PropTypes may need to be adjusted as well.  I
will double check tonight and let you know.

ynejati:
Okay, I think we are good now with my latest commit.

The new output definition for PropTypes now matches the old definition.

**propTypes.d.ts** (**new**)
image

**index.d.ts** (**old**)
image

FredyC:
Thanks!

PR #829, opened 2020-02-06
Fix #806

### LINKED ISSUE #806: observableRequiresReaction + mobx-react = Observable reactive state being read outside a reactive context
```typescript
import { render as renderDom } from 'react-dom';
import * as React from 'react';
import { observer } from 'mobx-react';
import { observable, configure, action, reaction } from 'mobx';

configure({
	computedRequiresReaction: true,
	observableRequiresReaction: true,
	reactionRequiresObservable: true,
	enforceActions: 'observed',
});

const globalStateInAnotherFile = observable({
	counter: 0,
});

@observer  /// <--------- THIS
class SimpleTest extends React.Component<any> {
	render() {
		return (
			<React.Fragment>
				<h1>{globalStateInAnotherFile.counter}</h1>
				<button onClick={() => this.testAction()}>i++</button>
			</React.Fragment>
		);
	}

	@action
	testAction() {
		globalStateInAnotherFile.counter++;
	}
}

reaction(
	() => globalStateInAnotherFile.counter,
	(v) => console.log('current value is %d', v)
);

renderDom(<SimpleTest />, document.getElementById('applicationRoot'));
```

On page render, console warn this 8 times:
```text
[mobx] Observable reactive state being read outside a reactive context 
    in SimpleTest
```

call stack show 5 different place causing the warning. all comes from `react-dom.development.js`, like:
* constructClassInstance: `var state = workInProgress.memoizedState = instance.state !== null && instance.state !== undefined ? instance.state : null;`
* checkClassInstance: `var _state = instance.state;`
* updateClassComponent: `if (inst.props !== nextProps) {`

Whenever I click the button, log:
```text
current value is 6
[mobx] Observable reactive state being read outside a reactive context 
    in SimpleTest
[mobx] Observable reactive props being read outside a reactive context 
    in SimpleTest
```

I can't fix this by reading the doc. Please help.

Bnaya:


src/observerClass.ts
```diff
@@ -153,7 +153,9 @@ function makeObservableProp(target: any, propName: string): void {
         configurable: true,
         enumerable: true,
         get: function() {
+            const prevReadState = _allowStateReadsStart(true);
```
mweststrate: I'd still null check these, just in case people didn't update their peers

test/issue806.test.js
```diff
@@ -0,0 +1,35 @@
...
+    configure({
+        observableRequiresReaction: true
+    });
+
+    const x = observable({
+        a: 1
+    });
+
+    withConsole(["warn"], () => {
+        render(<Issue806Component />);
```
mweststrate: I'd expand the test case to where one observer component passes some of it's observables to another observer component

src/observerClass.ts
```diff
@@ -153,7 +153,17 @@ function makeObservableProp(target: any, propName: string): void {
         configurable: true,
         enumerable: true,
         get: function() {
+            let prevReadState = false;
+
+            if (_allowStateReadsStart && _allowStateReadsEnd) {
```
Bnaya: @mweststrate should we `import * as mobx` or this null check is enough?

test/issue806.test.js
```diff
@@ -0,0 +1,45 @@
...
+import { render } from "@testing-library/react"
+import withConsole from "./utils/withConsole";
+
+
+@observer
+class Issue806Component extends React.Component {
+    render() {
+        return <span>
+            {this.props.a}
+            <Issue806Component2 propA={this.props.a} propB={this.props.b} />
```
Bnaya: I'm not sure why it was needed, but i've added these 😅

src/observerClass.ts
mweststrate: nope. this should work fine

FredyC:
I don't follow why to bump peer dep. It's not like it will force a consumer to install that specific version, only warn about it.

Besides, requiring a specific version feels like sort of a breaking change to me. The code has to work with X.0.0 versions.

mweststrate:
> I don't follow why to bump peer dep. It's not like it will force a consumer to install that specific version, only warn about it.

Yeah that is the kind of the point, if you bump mobx-react we want ideally want folks to bump mobx is as well to not run into #806 

> Besides, requiring a specific version feels like sort of a breaking change to me. The code has to work with X.0.0 versions.

It still does, but in that case you will still suffer from the false positive from #806

Bnaya:
Btw i think we have broken code. in the past,
Weve added new apis to mobx, used them without null checks in mobx-react, and without peer bump.
I woudnt mind revert the change and maybe add null checks + dev time warnings

FredyC:
It's already published, but feel free to improve for the next version.

PR #835, opened 2020-02-10
Transform template literals to fix IE11 compatibility

samjetski:
Hi, thanks for publishing this lib!

It appears that since the build tools overhaul in #821 all of the js bundles in `dist`/npm now include template literals (backticks ` `` ` with interpolation), an ES6 feature which effectively [breaks support for Internet Explorer 11 (and older)](https://caniuse.com/#search=template%20literals).

`dist/mobx-react.module.js:14` in v6.1.5:
```js
  var symbol = "__$mobx-react " + name + " (" + symbolId + ")";
```

`dist/mobxreact.cjs.development.js:20` in v6.1.7
```ts
  const symbol = `__$mobx-react ${name} (${symbolId})`;
```

The [symbol generation util](https://github.com/mobxjs/mobx-react/blob/bba25e367fe90953d12a4805f120fc58ea7fd505/src/utils/utils.ts#L6) appears to be the source, but it seems to be only have propagated this way to NPM in v6.1.6 onwards.

Adding [`@babel/plugin-transform-template-literals`](https://babeljs.io/docs/en/babel-plugin-transform-template-literals) to `.babelrc` appears to do the trick but I'm not entirely across the build tooling or other package distribution issues.

`dist/mobxreact.cjs.development.js:20` with this change:
```ts
  const symbol = "__$mobx-react ".concat(name, " (").concat(symbolId, ")");
```

Is this the right path forward? Cheers!

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,0 @@
-{
-    "presets": [["@babel/preset-env", { "targets": { "node": "current" } }]],
-    "plugins": [
-        ["@babel/plugin-proposal-decorators", { "legacy": true}],
-        ["@babel/plugin-proposal-class-properties", { "loose": true}],
-        "@babel/plugin-transform-react-jsx"
-    ]
-}
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -21,7 +21,7 @@
         "sourceMap": false,
         "strict": true,
         "suppressImplicitAnyIndexErrors": true,
-        "target": "es6"
+        "target": "es5"
     },
     "include": ["src/**/*.ts", "test/**/*.ts"]
 }


FredyC:
Well, I don't think Babel plugin will do the trick here. We are using TSDX since the latest patch version which means the code is built with TypeScript Compiler. The Babel remnants can be removed as they are not used AFAIK.

What needs to change is `target = es5` in `tsconfig.json`

samjetski:
Looks like TSDX does [use the existing .babelrc](https://github.com/jaredpalmer/tsdx#babel) and merges it, which is why adding this plugin has the desired effect. Worth keeping in mind when it comes to pulling out the old build config.

Ah you're right, `"target": "es5"` makes more sense and would certainly be cleaner, but it doesn't seem to work and I haven't figured out why - I'm still getting template literals in my `dist`. Any ideas?

samjetski:
I think I've got it, removing the `.babelrc` seemed to do the trick. I'm guessing this line is the culprit as it may have conflicted with the tsconfig.json `target`:
```js
"presets": [["@babel/preset-env", { "targets": { "node": "current" } }]]
```
Here's the new output for `dist/mobxreact.cjs.development.js:20`:
```ts
  var symbol = "__$mobx-react " + name + " (" + symbolId + ")";
```
Would appreciate some others verifying the build output as I don't know what other impacts this change may have.

FredyC:
You are probably right, have missed this part of TSDX using Babel sometimes. I will have a deeper look later afternoon.

Also looks like flipping target breaks tests, so I suppose we need different target there (tsconfig.test.json)

FredyC:
So it seems that using Babel would be a better option after all. Let's keep changes at a minimum. Can you please revert to your previous commit by adding that plugin?

asapach:
In the meantime could you please either point the `@latest` to `v6.1.5` on npm, or roll back the change and release a new version and/or [deprecate](https://docs.npmjs.com/cli/deprecate) affected package versions?

OscarBarrett:
FYI the 6.1.6/6.1.7 changes also break compatibility with Edge 17 which is less than 2 years old, as it doesn't support destructuring an object with the rest parameter.

mweststrate:
Published 6.1.8. Sorry for the late follow up, this is kinda embarrassing! @OscarBarrett  / @asapach / @samjetski I did a manual check that those language constructions are no longer used, could you verify things are running again in IE 11? If this last change doesn't help, I'm fine rolling back to 6.1.5 till we sort this out better, but I think we should be good now.

(If somebody knows a tool to statically verify that that a javascript file comforms to a certain version of the language spec, please tell me!) 

@FredyC let's make sure we follow up on this kind of fallout from recently released things asap :) This is breaking for a lot of people and probably hard to catch for many as well (something something with IE 11 and CI in general :/ )

asapach:
> If somebody knows a tool to statically verify that that a javascript file comforms to a certain version of the language spec, please tell me!

You can use [eslint for this](https://eslint.org/docs/user-guide/configuring#specifying-parser-options):
```json
{ "parserOptions": { "ecmaVersion": 5 } }
```

FredyC:
@asapach Thanks for the tip, although we want to keep source code in ES2015+. It would require running a linter for built files and probably disable all styling rules. It might be worth investigating for sure.

asapach:
@FredyC, you can use separate eslintc files for `src` and `dist`.

mweststrate:
note that that in itself is not enough yet, we should have some lint rules
to forbid (unguarded) usage for things like Symbol etc

Issue #838, opened 2020-02-14
V6 Question about PureComponents


creambyemute:
Hey there,

We just upgraded from 5.4.4 to 6.1.8 and I checked the changelogs to see if we have to adapt something.

We're still using Inject, planning on refactoring this for the new recommended way.

Question:
On the Changelog for 6.0.0 there is a mention of `Using PureComponent is now recommended for class-based components or React.memo for functional ones.`.

Does this mean, that if I have a observer Class Based Component `SearchIconBase extends React.Component<Props>` `inject("rootStore")(observer(SearchIconBase))`, that we should switch to `extends React.PureComponent<Props>`?

FredyC:
Uhh, we should probably expand that message. I know we were changing it once already but seeing it after a few months now it doesn't make any sense to me either :)

The `observer` will monkey patch our `shouldComponentUpdate` implementation if `PureComponent` is not used. I am not entirely sure about differences there, but yea, `PureComponent` is probably best.

https://github.com/mobxjs/mobx-react/blob/7d18dfdffff397dcb47f520f7f68a889dc24400e/src/observerClass.ts#L133

The `observer` for functional components will apply `React.memo` automatically.

creambyemute:
Thx for the response, so in theory, we should be switching to PureComponent.

It's a small change, I tried it yesterday and didn't see any side-effects.

FredyC:
Yea, PureComponent & friends are always about optimization, it's unlikely it would break anything unless your app code relies on optimization (bad practice imo).

I will keep this open and we should clarify the changelog further.

FredyC:
Related PR which improved readme regarding that, but kept the weird changelog statement.

https://github.com/mobxjs/mobx-react/pull/701

ynejati:
I ran into the same thing.  Ended up replacing the use of inject and creating our own provider using the context API.  None of the class based components made use of shouldComponent update, except for internal, non observer wrapper components.  As long as you don't use shouldComponentUpdate, which I think was always discouraged, you should be fine.

creambyemute:
We never used SCU, was totally fine to switch all of our 64 class components that were observers to be PureComponents now.

FredyC:
The inject/context has nothing to do with a problem. All related logic is within `observer`.

Using `PureComponent` is definitely better, but keep in mind it doesn't always work well for optimization purposes simply because `children` prop as well and will cause re-render anyway no matter if other props stay the same. Unless you memoize it in a parent of course.

[...]

PR #842, opened 2020-02-24
Issue #838: Clarify use of PureComponent and memo in changelog

ynejati:
Updated the changelog in response to #838.  Should we update the README as well in better detail?
  
-   [x] Updated changelog
-   [x] Updated `README` .....if applicable?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@
 **Improvements**
 
 -   Hook based components are now supported by mobx-react (in fact, the package is now implemented using hooks)
--   Class based `observer` components are now _recommended_ to extend `React.PureComponent`. Functional `observer` components are now automatically wrapped in `React.memo` internally. See section in [README](https://github.com/mobxjs/mobx-react/blob/master/README.md#observercomponentclass) for more details.
+-   Class based `observer` components are now _recommended_ to extend `React.PureComponent`. Functional `observer` components are now automatically wrapped in `React.memo` internally. See section in [README](https://mobx.js.org/README.html#observercomponentclass) for more details.
 -   For `observer` based components, there will now be an additional `Observer` component in the tree.
 -   Two new hooks have been exposed, in case you want to manage local state in observable: `useLocalStore` and `useAsObservableSource`.
 -   `MobXProviderContext` is now exposed from the package, in case you want to consume the context used by `Provider` with a `useContext` hook.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@
 **Improvements**
 
 -   Hook based components are now supported by mobx-react (in fact, the package is now implemented using hooks)
--   Class based `observer` components are now _recommended_ to extend `React.PureComponent`. Functional `observer` components are now _recommended_ to use `React.memo`. Doing so will help effectively communicate expected behaviour. If not followed, we apply patches for both internally so that the rules of `React.PureComponent` and `React.memo` are enforced.
+-   Class based `observer` components are now _recommended_ to extend `React.PureComponent`. Functional `observer` components are now automatically wrapped in `React.memo` internally. See section in [README](https://github.com/mobxjs/mobx-react/blob/master/README.md#observercomponentclass) for more details.
 -   For `observer` based components, there will now be an additional `Observer` component in the tree.
 -   Two new hooks have been exposed, in case you want to manage local state in observable: `useLocalStore` and `useAsObservableSource`.
 -   `MobXProviderContext` is now exposed from the package, in case you want to consume the context used by `Provider` with a `useContext` hook.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@
 **Improvements**
 
 -   Hook based components are now supported by mobx-react (in fact, the package is now implemented using hooks)
--   Using `PureComponent` is now _recommended_ for class-based components or `React.memo` for functional ones.
+-   Class based `observer` components are now _recommended_ to extend `React.PureComponent`. Functional `observer` components are now _recommended_ to use `React.memo`. Doing so will help effectively communicate expected behaviour. If not followed, we apply patches for both internally so that the rules of `React.PureComponent` and `React.memo` are enforced.
 -   For `observer` based components, there will now be an additional `Observer` component in the tree.
 -   Two new hooks have been exposed, in case you want to manage local state in observable: `useLocalStore` and `useAsObservableSource`.
 -   `MobXProviderContext` is now exposed from the package, in case you want to consume the context used by `Provider` with a `useContext` hook.
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Function (and decorator) that converts a React component definition, React compo
 
 When using component classes, `this.props` and `this.state` will be made observables, so the component will react to all changes in props and state that are used by `render`.
 
-`shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. The `observer` will automatically patch none pure class components with an internal implementation of `React.PureComponent` if necessary.
+`shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. The `observer` will automatically patch non-pure class components with an internal implementation of `React.PureComponent` if necessary.
 
 See the [MobX](https://mobxjs.github.io/mobx/refguide/observer-component.html) documentation for more details.
 
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Function (and decorator) that converts a React component definition, React compo
 
 When using component classes, `this.props` and `this.state` will be made observables, so the component will react to all changes in props and state that are used by `render`.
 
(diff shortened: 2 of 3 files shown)


FredyC:
Well, that even adds more confusion :) I mean class components are ok, but functional ones have `React.memo` inside and you don't need to worry about it. 

Anyway, I think it would be generally better to have more elaborate explanation in README and link that in changelog instead of copying it over.

ynejati:
Oops. You're right.  My fault.  I'll make it more brief and then just link to the updated README.

README.md
```diff
@@ -51,12 +51,19 @@ For greenfield projects you might want to consider to use [mobx-react-lite](http
 
 Please check [mobx.js.org](https://mobx.js.org) for the general documentation. The documentation below highlights some specifics.
 
-### `observer(componentClass)`
+### `observer(component)`
 
-Function (and decorator) that converts a React component definition, React component class or stand-alone render function into a reactive component, which tracks which observables are used by `render` and automatically re-renders the component when one of these values changes.
```
ynejati: The part that says "used by `render`" no longer feels relevant to functional components, so instead I said "used by its effective `render`".  Still doesn't seem right and might need to elaborate.

README.md
```diff
@@ -51,12 +51,19 @@ For greenfield projects you might want to consider to use [mobx-react-lite](http
...
+Function (and decorator) that converts a React component definition, React component class, or stand-alone render function, into a reactive component. A converted component will track which observables are used by its effective `render` and automatically re-render the component when one of these values changes.
+
+#### Functional Components
```
ynejati: Broke up into Functional and Class components sections because rules regarding one could easily get lost in the rules of another.

src/observer.tsx
```diff
@@ -27,7 +27,7 @@ export function observer<T extends IReactComponent>(component: T): T {
 
     if (ReactMemoSymbol && component["$$typeof"] === ReactMemoSymbol) {
         throw new Error(
-            "Mobx observer: You are trying to use 'observer' on function component wrapped to either another observer or 'React.memo'. The observer already applies 'React.memo' for you."
+            "Mobx observer: You are trying to use 'observer' on a function component wrapped in either another observer or 'React.memo'. The observer already applies 'React.memo' for you."
```
ynejati: Small changes here.  Easier to read.

CHANGELOG.md
```diff
@@ -74,7 +74,7 @@
 **Improvements**
 
 -   Hook based components are now supported by mobx-react (in fact, the package is now implemented using hooks)
--   Using `PureComponent` is now _recommended_ for class-based components or `React.memo` for functional ones.
+-   Class based `observer` components are now _recommended_ to extend `React.PureComponent`. Functional `observer` components are now automatically wrapped in `React.memo` internally. See section in [README](https://github.com/mobxjs/mobx-react/blob/master/README.md#observercomponentclass) for more details.
```
ynejati: Links to the README.

README.md
```diff
@@ -51,12 +51,19 @@ For greenfield projects you might want to consider to use [mobx-react-lite](http
...
+
+`React.memo` is automatically applied to functional components provided to `observer`. `Observer` does not accept a functional component already wrapped in `React.memo`, or an `observer`, in order to avoid consequences that might arise as a result of wrapping it twice.
+
+#### Class Components
 
 When using component classes, `this.props` and `this.state` will be made observables, so the component will react to all changes in props and state that are used by `render`.
-Note that `observer` automatically applies `React.memo` to any functional component you pass to it, however class components should extend `PureComponent` instead of `Component`
+
+`ShouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. `Observer` will automatically patch none pure class components with an internal implementation of `React.PureComponent` if necessary.
```
vkrol:
```suggestion
`shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. `Observer` will automatically patch none pure class components with an internal implementation of `React.PureComponent` if necessary.
```

README.md
```diff
@@ -51,12 +51,19 @@ For greenfield projects you might want to consider to use [mobx-react-lite](http
...
 
-### `observer(componentClass)`
+### `observer(component)`
 
-Function (and decorator) that converts a React component definition, React component class or stand-alone render function into a reactive component, which tracks which observables are used by `render` and automatically re-renders the component when one of these values changes.
+Function (and decorator) that converts a React component definition, React component class, or stand-alone render function, into a reactive component. A converted component will track which observables are used by its effective `render` and automatically re-render the component when one of these values changes.
+
+#### Functional Components
+
+`React.memo` is automatically applied to functional components provided to `observer`. `Observer` does not accept a functional component already wrapped in `React.memo`, or an `observer`, in order to avoid consequences that might arise as a result of wrapping it twice.
```
FredyC: Why would `Observer` have a problem with `React.memo` component? It does use a pure `useObserver` inside so it doesn't apply `React.memo`.

README.md
FredyC: I don't get second part of the paragraph. `Observer` is just re-export from mobx-react-lite and it doesn't do anything special to class components.

README.md
FredyC: I think it's fine like that.

CHANGELOG.md
FredyC:
```suggestion
-   Class based `observer` components are now _recommended_ to extend `React.PureComponent`. Functional `observer` components are now automatically wrapped in `React.memo` internally. See section in [README](https://mobx.js.org/README.html#observercomponentclass) for more details.
```

README.md
ynejati: It should be lower case.  I can see how that would be confused with the lite api.

FredyC: That wasn't my concern, I should have probably commented separately. Please reread why I said :)

README.md
ynejati: Again, should be lower case.  I'm referring to your logger message:

> "Mobx observer: You are trying to use 'observer' on a function component wrapped in either another observer or 'React.memo'. The observer already applies 'React.memo' for you."

FredyC: Ah, that makes sense then :) Yea, it's a bit unfortunate to have two things which differ in first letter casing but have such a different behavior.

README.md
```diff
@@ -57,13 +57,13 @@ Function (and decorator) that converts a React component definition, React compo
 
 #### Functional Components
 
-`React.memo` is automatically applied to functional components provided to `observer`. `Observer` does not accept a functional component already wrapped in `React.memo`, or an `observer`, in order to avoid consequences that might arise as a result of wrapping it twice.
+`React.memo` is automatically applied to functional components provided to `observer`. `observer` does not accept a functional component already wrapped in `React.memo`, or an `observer`, in order to avoid consequences that might arise as a result of wrapping it twice.
```
FredyC: Just for readability, too many observers imo :)
```suggestion
`React.memo` is automatically applied to functional components provided to `observer`. It does not accept a functional component already wrapped in `React.memo`, or an `observer`, in order to avoid consequences that might arise as a result of wrapping it twice.
```

README.md
```diff
@@ -57,13 +57,13 @@ Function (and decorator) that converts a React component definition, React compo
...
 
 #### Class Components
 
 When using component classes, `this.props` and `this.state` will be made observables, so the component will react to all changes in props and state that are used by `render`.
 
-`ShouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. `Observer` will automatically patch none pure class components with an internal implementation of `React.PureComponent` if necessary.
+`shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. `observer` will automatically patch none pure class components with an internal implementation of `React.PureComponent` if necessary.
```
FredyC:
```suggestion
`shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. The `observer` will automatically patch none pure class components with an internal implementation of `React.PureComponent` if necessary.
```

README.md
ynejati: Yeah, I thought about that.  Using 'it' instead makes it seem like you're talking about React.memo as it is the subject of the previous sentence.  And, rearranging the sentence could add confusion.  It's redundant, but it's clear, which is what I was aiming for.

README.md
```diff
@@ -51,12 +51,19 @@ For greenfield projects you might want to consider to use [mobx-react-lite](http
...
+#### Functional Components
+
+`React.memo` is automatically applied to functional components provided to `observer`. `observer` does not accept a functional component already wrapped in `React.memo`, or an `observer`, in order to avoid consequences that might arise as a result of wrapping it twice.
+
+#### Class Components
 
 When using component classes, `this.props` and `this.state` will be made observables, so the component will react to all changes in props and state that are used by `render`.
-Note that `observer` automatically applies `React.memo` to any functional component you pass to it, however class components should extend `PureComponent` instead of `Component`
+
+`shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. The `observer` will automatically patch none pure class components with an internal implementation of `React.PureComponent` if necessary.
```
vkrol:
```suggestion
`shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. The `observer` will automatically patch non-pure class components with an internal implementation of `React.PureComponent` if necessary.
```

FredyC:
Thanks @ynejati

PR #861, opened 2020-05-03
improve docs wording for class component behavior

chetnashah:
improve docs wording for class component behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ Function (and decorator) that converts a React component definition, React compo
 
 #### Class Components
 
-When using component classes, `this.props` and `this.state` will be made observables, so the component will react to all changes in props and state that are used by `render`.
+When using component classes, `this.props` and `this.state` will work like the standard react way, so the component will react to all changes in props and updates in state that are used by `render`.
 
 `shouldComponentUpdate` is not supported. As such, it is recommended that class components extend `React.PureComponent`. The `observer` will automatically patch non-pure class components with an internal implementation of `React.PureComponent` if necessary.
 


README.md
```diff
@@ -61,7 +61,7 @@ Function (and decorator) that converts a React component definition, React compo
 
 #### Class Components
 
-When using component classes, `this.props` and `this.state` will be made observables, so the component will react to all changes in props and state that are used by `render`.
+When using component classes, `this.props` and `this.state` will work like the standard react way, so the component will react to all changes in props and updates in state that are used by `render`.
```
FredyC, line 64: That doesn't seem right. They are observable, just not expected to be mutated to cause re-render.

chetnashah: "observable" in context of mobx means ability to react to mutation, right?

FredyC: Not exactly, the `observable` is something like a "primitive" which receives mutations and passes them to interested parties - reactions. Or at least that's my shallow understand, someone could probably provide a better description.

The way you have made that change doesn't really say anything useful imo.

To be honest, I am not using class components at all, so I fairly unsure why this behavior exists. I mean it's indeed React standard behavior to re-render when props/state changes. I am unclear why MobX needs to make those observable. In functional components, nothing like that exists. Perhaps someone with better insight can shine a light on it here.

cc @mweststrate @urugator @ynejati

ynejati: That sentence could definitely use a bit more elaboration.  Try your edits again after reading this:

>`this.props` and this.state are made observable, just to make sure `@computed` fields that are defined inside the component, and which rely on state or props, re-compute if state or props change (otherwise the computed wouldn't update and become stale on props change, since props are not observable.  However, this solution is not without it's own problems: https://github.com/mobxjs/mobx-react/issues?utf8=%E2%9C%93&q=is%3Aissue+label%3Aobservable-props-or-not+

chetnashah: I will dig deeper for props use case.
For this.state use case - I made this example: https://codesandbox.io/s/n4rq35r300?file=/src/index.js

FredyC: Um, how is that a use case @chetnashah? You are not supposed to mutate state in timeout **ever**, instead, you should call `this.setState`.

FredyC: @chetnashah Were you able to understand the problem to describe it better in docs?

mweststrate:
Closing for inactivity

Issue #887, opened 2020-07-27
upgrade mobx-react version from 6.2.2 to 6.2.3，There is an error


zengqingzhuang:
![image](https://user-images.githubusercontent.com/12510482/88510834-658fc400-d016-11ea-9f90-0ead998ddb33.png)

FredyC:
Please create a small reproduction.

joelharkes:
Yes have the same issue:

![image](https://user-images.githubusercontent.com/2960938/88539685-fd3ee380-d011-11ea-916e-ee6cb6408730.png)


We provide same component twice, because we use it for different stores, now it starts throwing errors. its quite annoying i had to downgrade to keep it working.

joelharkes:
It would be nice to at least be able to disable this error/warning.

mweststrate:
"We provide same component twice, because we use it for different stores"

Why do you need to wrap a component twice in observer twice because you use
it for two different stores? Please provide some example code why the
warning wouldn't be correct in your case.

bigdrum:
The warning is fine. The real problem is getDisplayName (https://github.com/mobxjs/mobx-react/blob/376bff72d84ae8d342c55d1fb25ff81a223624d2/src/observerClass.ts#L26) is passed with a undefined value; and it is not properly handled.

FredyC:
We will publish a fix soon, it was merely an oversight.

mweststrate:
Just release 6.2.4. Thanks for reporting!

PR #888, opened 2020-07-27
Issue #887: already defined observer class component warning message …

ynejati:
…should get display name from 'target' and not 'this'

### Code change checklist

-   [x] Updated changelog


FredyC:
Do you mind updating the changelog as well?

FredyC:
Thanks!

